Issue #1044, opened 2015-01-08
Show useful infomation if user requests divisions for a month


equivalentideas:
for example if someone goes to:

https://theyvoteforyou.org.au/divisions/representatives/2014-06

It show a blank page rather than 404 or results.

keithpitty:
Fixed.

PR #1058, opened 2015-06-08
Show divisions for a month

keithpitty:
See https://github.com/openaustralia/publicwhip/issues/1044

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -481,3 +481,6 @@ DEPENDENCIES
   uglifier (>= 1.3.0)
   vcr
   webmock
+
+BUNDLED WITH
+   1.10.3
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -85,10 +85,8 @@ module ApplicationHelper
     end
   end
 
-  def formatted_month(month, include_nbsp = false)
-    return nil unless month
-    date = Date.parse("#{month}-01")
-    include_nbsp ? date.strftime("%B&nbsp;%Y").html_safe : date.strftime("%B %Y")
+  def formatted_month(month)
+    Date.parse("#{month}-01").strftime("%B %Y")
   end
 
   def formatted_date(date, include_nbsp = false)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -85,6 +85,12 @@ module ApplicationHelper
     end
   end
 
+  def formatted_month(month, include_nbsp = false)
+    return nil unless month
+    date = Date.parse("#{month}-01")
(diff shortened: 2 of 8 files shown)


keithpitty:
Here's an example of this working locally:

divisions they vote for you 2015-06-08 18-48-47

app/helpers/application_helper.rb
```diff
@@ -85,6 +85,12 @@ def fraction_to_percentage_display(fraction, options = {precision: 2, significan
     end
   end
 
+  def formatted_month(month, include_nbsp = false)
```
henare: Since the `include_nbsp` parameter is never used when this function is called anywhere in the code, I wouldn't include it because that simplifies things.

app/helpers/application_helper.rb
```diff
@@ -85,6 +85,12 @@ def fraction_to_percentage_display(fraction, options = {precision: 2, significan
...
+    return nil unless month
```
henare: Explicit returns made this smell to me and now I look at it, why is this check needed at all?

henare:
Yay, this is a feature I've personally wanted for ages :sparkles: thanks @keithpitty!

I've added a couple of comments, I hope they help. Let me know what you think.

Since you've been keen on tests I think the helper and model changes in this PR could be great candidates to add tests for. Does that sound like a good idea to you?

keithpitty:
@henare thanks for your feedback! Yes, I was planning on adding some unit tests. Will hopefully have a chance to work on them and address your comments some time this evening.

henare:
Super, sounds great @keithpitty.

I've also got a couple of tips you might like for your next PR:

If you rebase instead of merging before you push your branch then you don't get extra commits like  5d28856 in this PR.

I'd also strongly recommend creating a feature branch when working on stuff for a pull request. You then open a PR from that branch and it means that `master` on your repository can match upstream (i.e. `openaustralia/publicwhip`). This means your `master` branch doesn't get out of sync with upstream and can save you lots of hassles when creating future PRs.

keithpitty:
@henare sorry, didn't get a chance this evening. Hopefully tomorrow.

keithpitty:
@henare I've cleaned up the helper usage and added some unit tests. What do you think? Is it ready to :ship:?

Gemfile.lock
```diff
@@ -481,3 +481,6 @@ DEPENDENCIES
   uglifier (>= 1.3.0)
   vcr
   webmock
+
+BUNDLED WITH
```
henare: I don't think you meant to commit this. I'll let it fly this time since it seems to be working ;)

henare:
Lovely @keithpitty, let's :shipit:

henare:
Deployed :rocket: https://theyvoteforyou.org.au/divisions/representatives/2015-06

PR #1071, opened 2015-09-28
Ukraine changes so far

henare:
This PR contains the application changes done for Ukraine so far.

There's nothing in it that precludes it from being merged.

### DIFF
--- a/app/models/division.rb
+++ b/app/models/division.rb
@@ -167,18 +167,14 @@ class Division < ActiveRecord::Base
     last_edit.is_a?(PaperTrail::Version) ? User.find(last_edit.whodunnit) : last_edit.user
   end
 
-  def oa_debate_url
+  def debate_url
     case house
     when "representatives"
       "http://www.openaustralia.org.au/debates/?id=#{oa_debate_id}"
     when "senate"
       "http://www.openaustralia.org.au/senate/?id=#{oa_debate_id}"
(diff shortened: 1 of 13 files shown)


app/models/division.rb
```diff
@@ -173,6 +173,10 @@ def oa_debate_url
       "http://www.openaustralia.org.au/debates/?id=#{oa_debate_id}"
     when "senate"
       "http://www.openaustralia.org.au/senate/?id=#{oa_debate_id}"
+    # TODO: This is temporarily here to make division pages load for the Ukraine
+    # we should probably rename this method and do something to support multiple countries
+    when "rada"
+      nil
```
henare: This should be fixed before we merge this.

lib/data_loader/popolo.rb
```diff
@@ -0,0 +1,113 @@
...
+            division.date = DateTime.parse(v_e["start_date"]).strftime("%F")
+            division.number = v_e["identifier"]
+            division.house = "rada" # TODO: Remove hardcoded value
+            division.name = v_e["title"]
+            division.source_url = "TODO"
+            division.debate_url = "TODO"
+            division.motion = "TODO"
+            division.clock_time = DateTime.parse(v_e["start_date"]).strftime("%T")
+            division.source_gid = v_e["identifier"]
+            division.debate_gid = "TODO"
```
henare: These TODOs should be TODONE before we merge this.

Issue #1053, opened 2015-02-26
Blank page when you click rebellion figure link on OA


henare:
To reproduce:

* Visit http://www.openaustralia.org.au/senator/michael_ronaldson/victoria#votingrecord
* Click "Never rebels against their party in this parliament." [link](https://theyvoteforyou.org.au/mp.php?id=uk.org.publicwhip/member/100192#divisions) at bottom of section
* Get blank page (profile page, divisions anchor on TVFY expected)

henare:
Seems to be Senators only. Their "More on their full record" link is missing too.

novemberkilo:
@henare I think I am able to reproduce this locally - comparing what I see locally to the stack trace on production should help determine that my local is breaking for good reasons :)

Any chance of please pasting the stack trace from honeybadger? Thanks.

novemberkilo:
Further notes following from @henare's description on how to reproduce the issue:

The "Never rebels against their party in parliament" link is: 

      https://theyvoteforyou.org.au/mp.php?id=uk.org.publicwhip/member/100192#divisions

This does not load, neither does (swap `member` for `senator`): 

      https://theyvoteforyou.org.au/mp.php?id=uk.org.publicwhip/senator/100192#divisions

But swapping `member` for `lord` does (although it does not appear to anchor at #divisions): 

     https://theyvoteforyou.org.au/mp.php?id=uk.org.publicwhip/lord/100192#divisions

henare:
>Any chance of please pasting the stack trace from honeybadger? Thanks.

@novemberkilo there's no stacktrace, it 404s.

novemberkilo:
So based on what I see thus far, the issue is that the link is being incorrectly generated for senators. The link currently has an `id` parameter that looks like `uk.org.publicwhip/member/100192` but for senators it should be `uk.org.publicwhip/lord/100192`

Does this sound right @henare?

henare:
>Oh right ... see I thought it would be a 500 because I'm erroring out locally with something like:
>
> Division#aye_majority delegated to division_info.aye_majority, but division_info is nil
>
>Which might be about my local db still being incomplete? 

D'oh! Yeah you need to [build the caches](https://github.com/openaustralia/publicwhip/#australia-1) (eugh!). Run this and you should be hunky dory:

    rake application:cache:all_except_member_distances

If you have ideas about tidying up the doco so this is clearer for other developers that would be fantastic :musical_note: 

[...]

PR #1087, opened 2016-03-16
Fix redirects if member is a senator

novemberkilo:
See #1053 

- [x] Fix situation where the URL says `id=uk.org.publicwhip/member/100192` even though Member with id 100192 has `gid=uk.org.publicwhip/lord/100192`
- ~~Fix so that page renders at anchor `#divisions`~~

### DIFF
--- a/spec/requests/members_controller_spec.rb
+++ b/spec/requests/members_controller_spec.rb
@@ -57,6 +57,14 @@ describe MembersController, type: :request do
       end
 
       it { compare_static("/mp.php?mpn=Barnaby_Joyce&mpc=New_England&house=representatives") }
+
+      it "should redirect to the senator's page even if the id param incorrectly identifies a member as a senator and vice versa" do
+        # Barnaby has been set up above as a member and a senator.
+        # Let's refer to his senator record but incorrectly using `member` instead of `lord`
+        get "/mp.php?id=uk.org.publicwhip/member/100114"
+        expect(response.status).to eq 302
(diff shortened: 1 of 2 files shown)


spec/requests/members_controller_spec.rb
```diff
@@ -37,7 +37,7 @@
     it {compare_static("/mp.php?mpn=Christine_Milne&mpc=Tasmania&house=senate&dmp=1")}
 
     # Test free teller under Interesting Votes
-    it {compare_static("/mp.php?mpn=Roger_Price&mpc=Chifley&house=representatives")}
+     it {compare_static("/mp.php?mpn=Roger_Price&mpc=Chifley&house=representatives")}
```
henare: Was this supposed to be committed?

spec/requests/members_controller_spec.rb
```diff
@@ -57,6 +57,12 @@
       end
 
       it { compare_static("/mp.php?mpn=Barnaby_Joyce&mpc=New_England&house=representatives") }
+
+      it "should redirect to the senator's page even if the gid is incorrect" do
```
henare: Yay tests! :smile: 

I wonder if we shouldn't explain in the test description why rather than just saying, "incorrect". I can imagine seeing this in the future and wondering, "why on earth would you do that?"

henare:
:star: :star2: :star: :star2: :star: :star2:

PR #1115, opened 2017-02-01
Handled error of invalid date in divisions page

### LINKED ISSUE #1063: ArgumentError - invalid date
Should 404 instead. Example URL: https://theyvoteforyou.org.au/divisions/senate/2015-06-122
## Backtrace

```
line 235 of [PROJECT_ROOT]/app/helpers/divisions_helper.rb: parse
line 235 of [PROJECT_ROOT]/app/helpers/divisions_helper.rb: divisions_period
line 15 of [PROJECT_ROOT]/app/views/divisions/index.html.haml: block in _app_views_divisions_index_html_haml__168451221_90067030
```

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/37542/faults/13546046)

mario-areias:
Fixes #1063

### DIFF
--- a/app/controllers/divisions_controller.rb
+++ b/app/controllers/divisions_controller.rb
@@ -31,9 +31,9 @@ class DivisionsController < ApplicationController
         @year = params[:date]
       elsif params[:date] =~/^\d{4}-\d{2}$/
         @month = params[:date]
-      else
+      elsif params[:date]
         begin
-          @date = Date.parse(params[:date]) if params[:date]
+          @date = Date.parse(params[:date])
         rescue ArgumentError => e
           invalid_date = params[:date]
           Rails.logger.info "Invalid date #{invalid_date}"
--- a/app/controllers/divisions_controller.rb
+++ b/app/controllers/divisions_controller.rb
@@ -32,7 +32,13 @@ class DivisionsController < ApplicationController
       elsif params[:date] =~/^\d{4}-\d{2}$/
         @month = params[:date]
       else
-        @date = params[:date]
+        begin
+          @date = Date.parse(params[:date]) if params[:date]
+        rescue ArgumentError => e
+          invalid_date = params[:date]
+          Rails.logger.info "Invalid date #{invalid_date}"
+          render 'divisions/division_not_found', status: 404
+        end
       end
       # Set the year to the lastest we have data for if it's not set
       @year = @years.last if @rdisplay.nil? && @date.nil? && @month.nil? && @year.nil?
--- a/app/controllers/divisions_controller.rb
+++ b/app/controllers/divisions_controller.rb
@@ -36,7 +36,6 @@ class DivisionsController < ApplicationController
           @date = Date.parse(params[:date])
         rescue ArgumentError => e
           invalid_date = params[:date]
-          Rails.logger.info "Invalid date #{invalid_date}"
           render 'home/error_404', status: 404
         end
       end
--- a/app/controllers/divisions_controller.rb
+++ b/app/controllers/divisions_controller.rb
@@ -35,7 +35,6 @@ class DivisionsController < ApplicationController
         begin
           @date = Date.parse(params[:date])
         rescue ArgumentError => e
-          invalid_date = params[:date]
           render 'home/error_404', status: 404
         end
       end
--- a/app/controllers/divisions_controller.rb
+++ b/app/controllers/divisions_controller.rb
@@ -37,7 +37,7 @@ class DivisionsController < ApplicationController
         rescue ArgumentError => e
           invalid_date = params[:date]
           Rails.logger.info "Invalid date #{invalid_date}"
-          render 'divisions/division_not_found', status: 404
+          render 'home/error_404', status: 404
         end
       end
       # Set the year to the lastest we have data for if it's not set
--- a/app/views/divisions/division_not_found.html.haml
+++ b/app/views/divisions/division_not_found.html.haml
@@ -1,6 +1,0 @@
-- content_for :title, 'Division not found'
-%h1= yield :title
-
-No Division found. If you entered a date, please make
-sure it is in the right format (YYYY-MM-DD).  Or you can
-#{link_to "browse all divisions", divisions_path}.
--- a/app/views/divisions/division_not_found.html.haml
+++ b/app/views/divisions/division_not_found.html.haml
@@ -1,0 +1,6 @@
+- content_for :title, 'Division not found'
+%h1= yield :title
+
+No Division found. If you entered a date, please make
+sure it is in the right format (YYYY-MM-DD).  Or you can
+#{link_to "browse all divisions", divisions_path}.
--- a/spec/controllers/divisions_controller_spec.rb
+++ b/spec/controllers/divisions_controller_spec.rb
@@ -1,5 +1,56 @@
 require 'spec_helper'
 
 describe DivisionsController, :type => :controller do
+  describe "#index" do
+    before :each do
+      Division.delete_all
 
+      @division = Division.create(id: 5, name: "hey jude", date: Date.new(2016,12,25),
+      number: 1, house: "representatives", source_url: "", debate_url: "", motion: "",
+      source_gid: "", debate_gid: "")
+
+      @other_division = Division.create(id: 6, name: "beautiful boy", date: Date.new(2016,06,01),
+      number: 1, house: "representatives", source_url: "", debate_url: "", motion: "",
+      source_gid: "", debate_gid: "")
+    end
+
+    context "when request has no parameters" do
+      it "should return index page with all divisions" do
+        get :index
+
+        expect(response).to render_template "divisions/index"
+        expect(response.status).to be 200
+        expect(assigns(:divisions)).to eq([@division, @other_division,])
+      end
+    end
+
+    context "when request has a date parameter that exists in the database" do
+      it "should return index page with selected divisions" do
+        get :index, date: '2016-06-01', house: "representatives"
+
+        expect(response).to render_template "divisions/index"
+        expect(response.status).to be 200
+        expect(assigns(:divisions)).to eq([@other_division])
+      end
+    end
+
+    context "when request has a date parameter that is not in database" do
+      it "should return index page with empty divisions" do
+        get :index, date: '2017-02-02', house: "representatives"
+
+        expect(response).to render_template "divisions/index"
+        expect(response.status).to be 200
+        expect(assigns(:divisions)).to be_empty
+      end
+    end
+
+    context "when request has an invalid date as a parameter" do
+      it "should return generic 404 page" do
+        get :index, date: '2017-02-0222', house: "representatives"
+
+        expect(response).to render_template "home/error_404"
+        expect(response.status).to be 404
+      end
+    end
+  end
 end
--- a/spec/controllers/divisions_controller_spec.rb
+++ b/spec/controllers/divisions_controller_spec.rb
@@ -2,65 +2,18 @@ require 'spec_helper'
 
 describe DivisionsController, :type => :controller do
   describe "#index" do
-    before :each do
-      Division.delete_all
+    before { Division.delete_all }
 
-      @division = Division.create(id: 5, name: "hey jude", date: Date.new(2016,12,25),
-      number: 1, house: "representatives", source_url: "", debate_url: "", motion: "",
-      source_gid: "", debate_gid: "")
+    let!(:division)  { create(:division, date: Date.new(2016,12,25)) }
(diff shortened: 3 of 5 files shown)


mario-areias:
I had to move the validation from `divisions_helper.rb` to `divisions_controller.rb`. Because the `divisions_helper#divisions_period` is used only in the template `app/views/divisions/index.html.haml` so it wouldn't be possible to have a 404, AFAIK. Let me know if that's not the correct approach in this case.


I also tried to create a controller test, but I was not able to complete it. I got up to this point:

``` ruby
require 'spec_helper'
describe DivisionsController, :type => :controller do
  it 'returns divisions root page' do
    get :index

    response.should render_template :divisions
  end
end
```

And I got this error

```
1) DivisionsController returns divisions root page
     Failure/Error: @years = (Division.order(:date).first.date.year..Division.order(:date).last.date.year).to_a


 NoMethodError:
  undefined method `date' for nil:NilClass
     # ./app/controllers/divisions_controller.rb:26:in `index'
```

If you know how to finish this test, I would be more than happy to add more including the scenario I fixed.

app/controllers/divisions_controller.rb
```diff
@@ -32,7 +32,13 @@ def index
       elsif params[:date] =~/^\d{4}-\d{2}$/
         @month = params[:date]
       else
-        @date = params[:date]
+        begin
+          @date = Date.parse(params[:date]) if params[:date]
+        rescue ArgumentError => e
+          invalid_date = params[:date]
+          Rails.logger.info "Invalid date #{invalid_date}"
```
mario-areias: I am know sure about this line, for two reasons:

1) Not sure if this is an error worth logging

2) It definitely needs better wording. 

Let me know if you want me to change this.

spec/helpers/divisions_helper_spec.rb
```diff
@@ -56,7 +56,7 @@
 
     context "date specified" do
       before do
-        helper.instance_variable_set("@date", "2014-06-01")
+        helper.instance_variable_set("@date", Date.parse("2014-06-01"))
       end
 
       it "returns formatted month when present" do
```
equivalentideas: This test description is wrong, we should fix this while we're here.

app/views/divisions/division_not_found.html.haml
```diff
@@ -0,0 +1,6 @@
+- content_for :title, 'Division not found'
+%h1= yield :title
+
+No Division found. If you entered a date, please make
```
mario-areias: That's a good point. I used Division because that's what is displayed in the top menu, so I assumed people knew somehow. But it seems I am wrong 🙂. 

So far, we are using this page only for the invalid date scenario and I put a message in there to explain why. Anyway, I agree the wording is not very good. If you have any suggestions in how to improve it, that would be great.

app/controllers/divisions_controller.rb
```diff
@@ -31,8 +31,14 @@ def index
         @year = params[:date]
       elsif params[:date] =~/^\d{4}-\d{2}$/
         @month = params[:date]
-      else
-        @date = params[:date]
+      elsif params[:date]
```
mario-areias: I totally agree with it. But I am not familiar with the code and I like baby steps. That's why I just did a simple fix. I tried to add a test, but there was an error that I couldn't figure out how to fix it. 

If we fix that, I can add more tests and learn exactly what this controller is doing. Then the refactor tasks gets much easier.

app/views/divisions/division_not_found.html.haml
equivalentideas: The other option is to redirect to a generic 404 page, which includes helpful information for people to either go to the home page, search or correct the url. It looks like we don't have a 404 page at all at the moment!! Yikes.

Because this error is probably quite rare to hit, it's probably not worth the effort/maintenance of a specific 404. What do you recon?

mario-areias: Sure. I can create something like it. It won't be very helpful for this scenario, but as you said, it is an rare error anyway. Any ideas on the wording?

app/controllers/divisions_controller.rb
equivalentideas: > But I am not familiar with the code and I like baby steps. 

Totes :) I think that's what made me think this is a first step pointing towards a bigger win :)

So, looking at the test example you gave above, I think what's happening is that it's trying to get the Divisions page, but the code that finds the divisions is making an assumption that there are in fact divisions in the database. It's raising an error because there's no divisions being created in the test setup.

You can create divisions for this test using our [Division factory](). We use factory_girl for this http://www.rubydoc.info/gems/factory_girl/file/GETTING_STARTED.md#Using_factories . Something like `create(:division)` will make you a valid division. You can use a Times loop to make a bunch.

mario-areias: Nice! I will give it a go tonight.

app/views/divisions/division_not_found.html.haml
equivalentideas: Huh, actually we already have a [404 page](https://github.com/openaustralia/publicwhip/blob/master/app/views/home/error_404.html.haml), [route](https://github.com/openaustralia/publicwhip/blob/master/config/routes.rb#L208), and [action](https://github.com/openaustralia/publicwhip/blob/master/app/controllers/home_controller.rb#L53). So I recon just redirect to that, and then improving that page could be a new issue/PR.

spec/controllers/divisions_controller_spec.rb
```diff
@@ -1,5 +1,76 @@
 require 'spec_helper'
 
 describe DivisionsController, :type => :controller do
+  describe "#index" do
+    before :each do
+      Division.delete_all
 
+      @division = Division.create(id: 5, name: "hey jude", date: Date.new(2016,12,25),
```
mario-areias: I could not use `let` here because it is necessary to have at least one division in the database when doing a GET request in `/divisions`

equivalentideas: Could you use let with the !, `let!`? [It runs the statement before each example](https://relishapp.com/rspec/rspec-core/docs/helper-methods/let-and-let):

```
let!(:division1) { create(:division, date: Time.new(2016,12,25) }
let!(:division2) { create(:division, date: Time.new(2016,06,01) }
```

That also uses the factories so you only have to assign the data we care about for this test, and the factory sets the other stuff for a valid Division.

spec/controllers/divisions_controller_spec.rb
```diff
@@ -1,5 +1,76 @@
 require 'spec_helper'
 
 describe DivisionsController, :type => :controller do
+  describe "#index" do
+    before :each do
+      Division.delete_all
```
equivalentideas: Could you add to this a comment like in `spec/features/viewing_user_stats_spec.rb` to remind us to fix this crappy situation requiring this database clearing, e.g. `# TODO: Remove this hack to delete fixtures`?

I'd love to get around to fixing that one day :)

spec/controllers/divisions_controller_spec.rb
```diff
@@ -1,5 +1,76 @@
...
+
+        expect(response).to render_template "divisions/index"
+        expect(response.status).to be 200
+        expect(assigns(:divisions)).to eq([@other_division])
+      end
+    end
+
+    context "when request has a year parameter that exists in the database" do
+      it "should return index page with selected divisions" do
+        get :index, date: '2016', house: "representatives"
```
equivalentideas: Do we need to specify a house here? That seems irrelevant to this test.

spec/controllers/divisions_controller_spec.rb
```diff
@@ -1,5 +1,76 @@
...
+      number: 1, house: "representatives", source_url: "", debate_url: "", motion: "",
+      source_gid: "", debate_gid: "")
+
+      @other_division = Division.create(id: 6, name: "beautiful boy", date: Date.new(2016,06,01),
+      number: 1, house: "representatives", source_url: "", debate_url: "", motion: "",
+      source_gid: "", debate_gid: "")
+    end
+
+    context "when request has no parameters" do
+      it "should return index page with all divisions" do
```
equivalentideas: "with all divisions" I don't think this is actually true. If you load https://theyvoteforyou.org.au/divisions/ it seems to include divisions for the latest calendar year that we have divisions for.

spec/controllers/divisions_controller_spec.rb
```diff
@@ -1,5 +1,76 @@
...
+        get :index
+
+        expect(response).to render_template "divisions/index"
+        expect(response.status).to be 200
+        expect(assigns(:divisions)).to eq([@division, @other_division])
+      end
+    end
+
+    context "when request has a complete date parameter that exists in the database" do
```
equivalentideas: I think this wording could be clearer. Maybe "when the date parameter is a full date", it "renders the index template with only divisions that occurred on that date". This removes the mention of the database from the context, because it's not really relevant at that point.

Also I think it's better to talk about the "rendering" or "assigning" here than the "returning", because we're not actually testing what appears on the page. Returns just seems like the wrong word to me :) maybe it seems confusing with the idea of a method returning something. What do you recon?

spec/controllers/divisions_controller_spec.rb
```diff
@@ -1,5 +1,76 @@
 require 'spec_helper'
 
 describe DivisionsController, :type => :controller do
+  describe "#index" do
```
equivalentideas: Looking good ⭐️ It's great to start having this crazy action tested. I've added some suggestions. See what you think.

With those tests in place I'd be happy to merge this.

spec/controllers/divisions_controller_spec.rb
mario-areias: Awesome, I didn' t know about `let!`. I tried create divisions like you suggested, but this factories creates way too many divisions. I will try again now, but I am afraid I will still need to use `Divisions.create'`

spec/controllers/divisions_controller_spec.rb
mario-areias: That's accurate. I noticed after wrote this test and forgot to change. Will rephrase it.

spec/controllers/divisions_controller_spec.rb
mario-areias: Hum. The database bit is relevant, if it is not in the database, you get a `404`. Agree about "returning" I will change for "rendering". My English skills are not great, so don't feel ashamed to ask me to reword something :)

spec/controllers/divisions_controller_spec.rb
mario-areias: You are right. It is not relevant. But that's how it works in production, you need to specify a house in order to choose a date. That's the route for this request: `/divisions/:house/:date/`. I like the idea of tests being a specification, that's why I added. I would like to leave it here, unless you feel strong about removing it :)

spec/controllers/divisions_controller_spec.rb
mario-areias: I was wrong about this  ^. I can do as you suggest \o/

spec/controllers/divisions_controller_spec.rb
mario-areias: Actually, this is not accurate. The code for that is this:

```ruby
@years = (Division.order(:date).first.date.year..Division.order(:date).last.date.year).to_a
# lots 
# of 
# code 
# here...

# Set the year to the lastest we have data for if it's not set
@year = @years.last if @rdisplay.nil? && @date.nil? && @month.nil? && @year.nil?
```

Which in production, it will happens as you say as long as we keep adding divisions. But the difference is important to do my testing, because that way I don't need to mock the System date :)

spec/controllers/divisions_controller_spec.rb
mario-areias: Ooops, you get a `200` but with empty divisions.

spec/controllers/divisions_controller_spec.rb
mario-areias: Done. I think I addressed most of your suggestions. I also used nested `describes`, I wasn't able to find out if that's a good practice in ruby, but it seems good to me. Let me know what you think. Also the search by `month` seems cheesy. I will open an issue about it.

equivalentideas:
> Done. I think I addressed most of your suggestions. I also used nested describes, I wasn't able to find out if that's a good practice in ruby, but it seems good to me. Let me know what you think :)

Looking good @mario-areias . There's a few little clean up things I'll do and I'll merge this. It's great that fixing this little bug and also setting things up for a bigger improvement :)

equivalentideas:
Merged :) Check out the clean up changes I made on the end here https://github.com/openaustralia/publicwhip/compare/e28bb0650c85...35292c8da616

PR #1118, opened 2017-02-10
Upgrade to ruby 2.3.1

CloCkWeRX:
Like we did for planningalerts, this jumps over to ruby 2.3.1


equivalentideas:
I've assigned @henare to this, he said he wanted to do it. Thanks for kicking it off @CloCkWeRX !

henare:
I've started to clean up deprecation and other warnings.

@CloCkWeRX feedback for doing this next time - when doing upgrades just getting it working isn't enough. You need to make it work as much as possible like it was before. This means addressing things like deprecation warnings or other changes in behaviour. It's important to do this because otherwise you build up bit-rot and unexpected behaviour in your application is bound to start happening. It also makes future upgrades much more difficult.

It's also essential to smoke test major parts of the application functionality. This is because automated tests can't catch everything and there's no substitute for seeing things working. I've upgraded an application before that had a totally green test suite that wouldn't even boot when you actually tried to run it! :smile:

CloCkWeRX:
> ... need to make it work as much as possible like it was before. This means addressing things like deprecation warnings or other changes in behaviour. 

Its not that I ignore deprecation warnings, it just that I hate sending pull requests that touch "everything" and may actually hide regression; if I can integrate a small change (with no negative behaviour changes); deploy/deliver; then integrate a larger change all of one type, repeat.

Take a peek at what we do on https://github.com/Growstuff/growstuff/pulls?q=is%3Apr+is%3Aclosed - tech debt is pretty viciously stamped out; and it tends to make it easy to bisect/revert/remove problems.

> It's also essential to smoke test major parts of the application functionality. 

I unfortunately don't have the local resources to effectively smoke test (~12 medium sized apps for work, maintining ffcrm, growstuff and contributing to openfarm); nor the in depth experience to do it properly.

Run app and get a hello world, sure (that's already covered by your feature tests); but end to end UAT style testing... :s not so much.

I know that its time consuming to onboard contributions like this; I'm more than happy to contribute feature tests if that'll help push things down a layer in the test pyramid; or add other tooling to catch problems - I just need a bit of guidance on where to put the effort.

henare:
@CloCkWeRX I appreciate your response - that all make sense to me :+1:

PR #1122, opened 2017-03-02
Refactor date validation on division_controller_spec

mario-areias:
This is to finish the work around date validation on the divisions controller page I have started in January. Now it always returns 404 if date is invalid (i.e. 2017-13-12) or date has a bad format (i.e. 2017-111-20). 

I also make an extra effort to make sure the tests comparing the generated pages with the legacy php pages at `spec/requests/divisions_controller_spec.rb` didn't break. Not sure how important is to keep this compatibility, but for now it seems to work good.

### DIFF
--- a/app/controllers/divisions_controller.rb
+++ b/app/controllers/divisions_controller.rb
@@ -27,19 +27,14 @@ class DivisionsController < ApplicationController
       @sort = params[:sort]
       @rdisplay = params[:rdisplay]
       @house = params[:house] unless params[:house] == "all"
-      if params[:date] =~ /^\d{4}$/
-        @year = params[:date]
-      elsif params[:date] =~/^\d{4}-\d{2}$/
-        @month = params[:date]
-      elsif params[:date]
-        begin
-          @date = Date.parse(params[:date])
-        rescue ArgumentError => e
-          render 'home/error_404', status: 404
-        end
+
+      begin
+        @date_start, @date_end, @date_range = DivisionParameterParser.get_date_range(params[:date]) if params[:date]
+        @date_start, @date_end, @date_range = DivisionParameterParser.get_date_range(@years.last.to_s) if @date_start.nil? && @date_end.nil? && @rdisplay.nil?
+      rescue ArgumentError => e
+        render 'home/error_404', status: 404
       end
-      # Set the year to the lastest we have data for if it's not set
-      @year = @years.last if @rdisplay.nil? && @date.nil? && @month.nil? && @year.nil?
+
       # This sets the parliament to display if it's not set. It's only here for legacy support
       # and should probably be cleaned up at some stage as we no longer focus on parliament sessions
       @rdisplay = "2013" if @rdisplay.nil?
@@ -77,10 +72,8 @@ class DivisionsController < ApplicationController
       @divisions = Division.order(order)
       @divisions = @divisions.joins(:division_info) if @sort == "rebellions" || @sort == "turnout"
       @divisions = @divisions.in_house(@house) if @house
-      @divisions = @divisions.on_date(@date) if @date
-      @divisions = @divisions.in_month(@month) if @month
-      @divisions = @divisions.in_year(@year) if @year
-      @divisions = @divisions.in_parliament(Parliament.all[@rdisplay]) unless @rdisplay == "all" || @date || @year
+      @divisions = @divisions.on_date_range(@date_start, @date_end) if @date_start && @date_end
+      @divisions = @divisions.in_parliament(Parliament.all[@rdisplay]) unless @rdisplay == "all" || @date_start || @date_end
       @divisions = @divisions.joins(:whips).where(whips: {party: @party}) if @party
       @divisions = @divisions.includes(:division_info, :wiki_motions, :whips)
     end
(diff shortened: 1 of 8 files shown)


app/controllers/divisions_controller.rb
```diff
@@ -27,19 +27,14 @@ def index
...
-        begin
-          @date = Date.parse(params[:date])
-        rescue ArgumentError => e
-          render 'home/error_404', status: 404
-        end
+
+      begin
+        @date_start, @date_end, @date_range = DivisionParameterParser.get_date_range(params[:date]) if params[:date]
+        @date_start, @date_end, @date_range = DivisionParameterParser.get_date_range(@years.last.to_s) if @date_start.nil? && @date_end.nil? && @rdisplay.nil?
+      rescue ArgumentError => e
```
henare: I know this is just carried over from above but I've realised we don't use the assigned `e` variable so we can make this just:

```ruby
rescue ArgumentError
```

app/models/division.rb
```diff
@@ -10,9 +10,7 @@ class Division < ActiveRecord::Base
 
   delegate :turnout, :aye_majority, :rebellions, :majority, :majority_fraction, to: :division_info
 
-  scope :on_date, ->(date) { where(date: date) }
-  scope :in_month, ->(month) { where("date >= ? AND date < ?", "#{month}-01", next_month(month)) }
-  scope :in_year, ->(year) { where("date >= ? AND date <= ?", "#{year}-01-01", "#{year}-12-31") }
+  scope :on_date_range, ->(date_start, date_end) { where("date >= ? AND date < ?", date_start, date_end) }
```
mario-areias: Yeap, that makes sense. Will change it.

app/models/division_parameter_parser.rb
```diff
@@ -0,0 +1,48 @@
+class DivisionParameterParser
+
+  YEAR_ONLY_REGEX = /^\d{4}$/
+  YEAR_AND_MONTH_REGEX = /^\d{4}-\d{1,2}$/
+  COMPLETE_DATE_REGEX = /^\d{4}-\d{1,2}-\d{1,2}$/
+
+  def self.get_date_range(date)
+
+    if date =~ YEAR_ONLY_REGEX
+      return get_year_range(date)
```
mario-areias: I thought so. But if I remove the return it doesn't work. I guess that's because the `get_year_range` method returns two values at the same time, but I am not quite sure :(

app/models/division_parameter_parser.rb
```diff
@@ -0,0 +1,48 @@
...
+      return get_day_range(date)
+    end
+
+    raise ArgumentError, 'Not a valid date format'
+  end
+
+private
+
+  def self.get_year_range(year)
+    date = Date.parse("#{year}-01-01")
```
mario-areias: I think semantically makes sense, because that's exactly what we are trying to do: parse a Date. Also, it seems the `.new` method does not accept a `String` as a parameter.

http://ruby-doc.org/stdlib-2.4.0/libdoc/date/rdoc/Date.html#method-c-new

app/models/division_parameter_parser.rb
```diff
@@ -0,0 +1,48 @@
...
+    year = date_parts.first
+    month = date_parts.last
+
+    date = Date.parse("#{year}-#{month}-01")
+
+    return date, date + 1.month, :month
+  end
+
+  def self.get_day_range(complete_date)
+    date_parts = complete_date.split('-')
```
mario-areias: That's neat! :heart: ruby :)

spec/controllers/divisions_controller_spec.rb
```diff
@@ -21,7 +21,16 @@
...
       it "should return generic 404 page" do
-        get :index, date: '2017-02-0222', house: "representatives"
+        get :index, date: '2017-13-22', house: "representatives"
+
+        expect(response).to render_template "home/error_404"
+        expect(response.status).to be 404
+      end
+    end
+
+    context "when request has an date parameter with an incorret format" do
```
mario-areias: Ooops, will fix

spec/models/division_parameter_parser_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'spec_helper'
+
+describe DivisionParameterParser, type: :model do
+  describe '::get_date_range' do
```
mario-areias: Alright, will fix. I thought it was the same notation used in the Ruby docs.

spec/models/division_parameter_parser_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    it 'gets range for complete date\'s format' do
+      date_start, date_end, date_range = DivisionParameterParser.get_date_range('2017-12-20')
+
+      expect(date_start).to eq(Date.parse('2017-12-20'))
```
mario-areias: Agreed. Will change it.

mario-areias:
@henare I pushed some changes and I believe this is ready to merge. Let me know if you want any other changes.

app/models/division_parameter_parser.rb
mario-areias: Actually, you are right. Here I can use implicit return. However, but in the `get_year_range` I can't because it returns more than 1 value.

app/models/division_parameter_parser.rb
henare: I think you're _constructing_ a date rather than parsing it here but I can't think of a compelling reason it needs to change so let's go with your choice :+1:

henare:
@mario-areias awesome - looks really tight :+1:

I found some more explicit returns to change and rather than flicking such a small thing back to you I wanted to add them here. I found I couldn't but I think I just had my password wrong :blush: I merged and then added the change as c32ea184c2bab7d43cf2e6c6b0b5169c63f13ee5.

Deployed :rocket:

PR #1128, opened 2017-03-09
Added pagination for MP page

### LINKED ISSUE #1123: All MP votes page is big and extremely slow
e.g. https://theyvoteforyou.org.au/people/representatives/cowper/luke_hartsuyker/divisions is 1.7 MB and takes over 60 seconds to load.

mario-areias:
I am not happy with this PR yet, but I would like some feedback on the solution (i.e. how the pagination was implemented). Will do some refactoring on this tomorrow night.

Resolves #1123

### DIFF
--- a/app/controllers/divisions_controller.rb
+++ b/app/controllers/divisions_controller.rb
@@ -33,6 +33,7 @@ class DivisionsController < ApplicationController
       if @member.nil?
         render 'members/member_not_found', status: 404
       else
+        @divisions = @member.divisions_possible(@date_start, @date_end).order(date: :desc, clock_time: :desc, name: :asc)
         render 'index_with_member'
       end
     else
--- a/app/controllers/divisions_controller.rb
+++ b/app/controllers/divisions_controller.rb
@@ -11,30 +11,23 @@ class DivisionsController < ApplicationController
     @years = (Division.order(:date).first.date.year..Division.order(:date).last.date.year).to_a
 
     if params[:mpc] && params[:mpn]
-      electorate = params[:mpc].gsub("_", " ")
-      name = params[:mpn].gsub("_", " ")
-
       @mpc = params[:mpc]
       @mpn = params[:mpn]
       @house = params[:house]
 
-      @member = Member.with_name(name)
-      @member = @member.in_house(@house)
-      @member = @member.where(constituency: electorate)
-      @member = @member.order(entered_house: :desc).first
+      @member = get_member
 
       begin
-        @date_start, @date_end = get_date_range(params[:date]) if params[:date]
-        @date_start, @date_end = get_date_range(@years.last.to_s) if @date_start.nil? && @date_end.nil?
+        @date_start, @date_end = get_date_range(params[:date])
       rescue ArgumentError
         render 'home/error_404', status: 404
       end
 
-      if @member.nil?
-        render 'members/member_not_found', status: 404
-      else
+      if @member
         @divisions = @member.divisions_possible(@date_start, @date_end).order(date: :desc, clock_time: :desc, name: :asc)
         render 'index_with_member'
+      else
+        render 'members/member_not_found', status: 404
       end
     else
       @sort = params[:sort]
@@ -42,8 +35,7 @@ class DivisionsController < ApplicationController
       @house = params[:house] unless params[:house] == "all"
 
       begin
-        @date_start, @date_end, @date_range = get_date_range(params[:date]) if params[:date]
-        @date_start, @date_end, @date_range = get_date_range(@years.last.to_s) if @date_start.nil? && @date_end.nil? && @rdisplay.nil?
+        @date_start, @date_end, @date_range = get_date_range(params[:date], @rdisplay)
       rescue ArgumentError
         render 'home/error_404', status: 404
       end
@@ -208,7 +200,23 @@ class DivisionsController < ApplicationController
     params.require(:policy_division).permit(:policy_id, :vote)
   end
 
-  def get_date_range(date)
-    DivisionParameterParser.get_date_range(date)
+  def get_date_range(date, rdisplay = nil)
+    if date
+      DivisionParameterParser.get_date_range(date)
+    elsif rdisplay.nil?
+      DivisionParameterParser.get_date_range(@years.last.to_s)
+    else
+      nil
+    end
+  end
+
+  def get_member
+    electorate = @mpc.gsub("_", " ")
+    name = @mpn.gsub("_", " ")
+
+    Member.with_name(name)
+    .in_house(@house)
+    .where(constituency: electorate)
+    .order(entered_house: :desc).first
   end
 end
--- a/app/controllers/divisions_controller.rb
+++ b/app/controllers/divisions_controller.rb
@@ -8,14 +8,27 @@ class DivisionsController < ApplicationController
   end
 
   def index
+    @years = (Division.order(:date).first.date.year..Division.order(:date).last.date.year).to_a
+
     if params[:mpc] && params[:mpn]
       electorate = params[:mpc].gsub("_", " ")
       name = params[:mpn].gsub("_", " ")
 
+      @mpc = params[:mpc]
+      @mpn = params[:mpn]
+      @house = params[:house]
+
       @member = Member.with_name(name)
-      @member = @member.in_house(params[:house])
+      @member = @member.in_house(@house)
       @member = @member.where(constituency: electorate)
       @member = @member.order(entered_house: :desc).first
+
+      begin
+        @date_start, @date_end = get_date_range(params[:date]) if params[:date]
+        @date_start, @date_end = get_date_range(@years.last.to_s) if @date_start.nil? && @date_end.nil?
+      rescue ArgumentError
+        render 'home/error_404', status: 404
+      end
 
       if @member.nil?
         render 'members/member_not_found', status: 404
@@ -23,14 +36,13 @@ class DivisionsController < ApplicationController
         render 'index_with_member'
       end
     else
-      @years = (Division.order(:date).first.date.year..Division.order(:date).last.date.year).to_a
       @sort = params[:sort]
       @rdisplay = params[:rdisplay]
       @house = params[:house] unless params[:house] == "all"
 
       begin
-        @date_start, @date_end, @date_range = DivisionParameterParser.get_date_range(params[:date]) if params[:date]
-        @date_start, @date_end, @date_range = DivisionParameterParser.get_date_range(@years.last.to_s) if @date_start.nil? && @date_end.nil? && @rdisplay.nil?
+        @date_start, @date_end, @date_range = get_date_range(params[:date]) if params[:date]
+        @date_start, @date_end, @date_range = get_date_range(@years.last.to_s) if @date_start.nil? && @date_end.nil? && @rdisplay.nil?
       rescue ArgumentError
         render 'home/error_404', status: 404
       end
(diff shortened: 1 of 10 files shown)


app/views/divisions/index_with_member.html.haml
```diff
@@ -9,5 +9,5 @@
 
 %p All votes #{@member.full_name_no_electorate} could have attended.
 
-- @member.person.members.order(entered_house: :desc).each do |member|
-  = render "divisions", member: @member, divisions: member.divisions_possible.order(date: :desc, clock_time: :desc, name: :asc)
+= render "divisions", member: @member, divisions: @member.divisions_possible(@date_start, @date_end).order(date: :desc, clock_time: :desc, name: :asc)
```
mario-areias: This line is actually from my other PR #1127. The relevant line for this PR is the below one.

spec/fixtures/static_pages/mp.php?mpn=Christine_Milne&mpc=Tasmania&house=senate&display=everyvote.html
```diff
@@ -106,14 +106,14 @@ <h2 class="social-share-heading small">
 
 <h2 id="divisions">All Votes</h2>
 <p>All votes Senator Christine Milne could have attended.</p>
-<ol class="display-house-all divisions-list list-unstyled">
+<ol class="display-house-senate divisions-list list-unstyled">
```
mario-areias: I am not sure what to do about these tests. Most changes seems reasonable, like deleting votes that are from different years and adding the pagination section. However, I am not sure why this class has changed or why the `span` below lost its text.

app/views/divisions/index_with_member.html.haml
```diff
@@ -9,5 +9,5 @@
 
 %p All votes #{@member.full_name_no_electorate} could have attended.
 
-- @member.person.members.order(entered_house: :desc).each do |member|
-  = render "divisions", member: @member, divisions: member.divisions_possible.order(date: :desc, clock_time: :desc, name: :asc)
+= render "divisions", member: @member, divisions: @divisions
```
mario-areias: It seems my comment here was deleted. But anyway, this line is from my previous PR #1127

app/controllers/divisions_controller.rb
```diff
@@ -8,29 +8,34 @@ def index_redirect
...
     else
-      @years = (Division.order(:date).first.date.year..Division.order(:date).last.date.year).to_a
       @sort = params[:sort]
       @rdisplay = params[:rdisplay]
       @house = params[:house] unless params[:house] == "all"
 
       begin
-        @date_start, @date_end, @date_range = DivisionParameterParser.get_date_range(params[:date]) if params[:date]
-        @date_start, @date_end, @date_range = DivisionParameterParser.get_date_range(@years.last.to_s) if @date_start.nil? && @date_end.nil? && @rdisplay.nil?
+        @date_start, @date_end, @date_range = get_date_range(params[:date], @rdisplay)
```
mario-areias: I am not sure what this `@rdisplay` does. But it seems something is wrong with this field. The `Parliament` class also looks pretty strange. Are we using this at all in production? Is that important or can be removed?

app/controllers/divisions_controller.rb
```diff
@@ -8,29 +8,34 @@ def index_redirect
...
-      name = params[:mpn].gsub("_", " ")
+      @mpc = params[:mpc]
+      @mpn = params[:mpn]
+      @house = params[:house]
 
-      @member = Member.with_name(name)
-      @member = @member.in_house(params[:house])
-      @member = @member.where(constituency: electorate)
-      @member = @member.order(entered_house: :desc).first
+      @member = get_member
```
mario-areias: I liked this specific block of refactoring, if people like it I will do similar refactorings with the rest of this controller

app/views/divisions/index_with_member.html.haml
equivalentideas: A cleaner way to achieve this would have been to rebase this brance on top of master so you're working from the latest changes. Doesn't create a merge conflict so all good.

spec/fixtures/static_pages/mp.php?mpn=Christine_Milne&mpc=Tasmania&house=senate&display=everyvote.html
equivalentideas: All these static html regression tests are to make sure that we're not accidentally changing the behaviour. The reason for this is that this app was converted over from the very old PHP code base. Some of the logic is very complex and there's lots of different stuff going on, and there's not a lot of unit tests. So the conversation was done using these regression tests to make sure views were being rendered in the same way.

So each change you see in these static pages is the result of some change you've made. It's all pretty complex, but you should track down where this change is coming from. It could be pointing to unintended side effects of the controller changes.

app/controllers/divisions_controller.rb
equivalentideas: I dig it too @mario-areias . Does it also feel like it's pointing toward some kind of simple MemberParser class or something like what you did here https://github.com/openaustralia/publicwhip/pull/1122/files ?  I haven't looked but I suspect we might be doing this kind of member look up in other controllers as well, so that could be a way to run the repetition.

I just thought of that looking at the attached controller tests, which feel like they could eventually be replaced with unit tests on the parsing stuff and an integration test of what the user is seeing.

What do you recon? Does that feel like a separate PR?

app/controllers/divisions_controller.rb
equivalentideas: > I am not sure what this @rdisplay does. 

Looking at https://github.com/openaustralia/publicwhip/blob/d27b8db198087f4b7b6b5f33905f6c8550c5b382/app/models/parliament.rb it appears to be a way of selecting different parliaments (a parliament is kinda a period of time, between to elections. So after each election they say, "looks like the 57th Parliament is going to be a cracker mate!" It's definitely insider jargon and should be explained in a comment on the model). Like I want to see divisions from the 2007 Parliament (2008-2010) https://theyvoteforyou.org.au/divisions?rdisplay=2007 what a confusing name "rdisplay"?

We don't expose that in the UI anywhere that I know of. We've gone for the more intuitive years in the pagination which I think suits people's mental model better. I can see how this could be useful though.

I'm not sure about removing it. It could be worth opening an issue for it. Alternatives could be to add it to the API docs or the UI somehow in a discrete way. Or rename it something useful so devs know what this is and add a comment to the Parliament model :) That feels like a separate PR.

Does that help?

spec/fixtures/static_pages/mp.php?mpn=Christine_Milne&mpc=Tasmania&house=senate&display=everyvote.html
mario-areias: Yeap. I am aware of that. What I was trying to say is that I tried to track down and I didn't manage to figure out why that changed. Sorry if I wasn't clear enough.

app/views/divisions/index_with_member.html.haml
mario-areias: Sure. But when I created this PR my other PR wasn't merged yet :smile:

app/controllers/divisions_controller.rb
mario-areias: It does help, definitely. Thanks for that. I will create another issue for that, although, I prefer to delete it. I like the idea to have as few features as possible. To make sure we're focusing in high value, user oriented features. If no one is using that or asking for that feature it's a smell that should be deleted. But anyway, I will add the issue and we can discuss more there.

app/controllers/divisions_controller.rb
mario-areias: Totally agree with you. After I finish this PR I can have a look in other controllers to verify if this is a problem elsewhere and then do a proper refactoring :smile:

mario-areias:
Sorry for the lack of context, @equivalentideas. I had the wrong assumption that @henare would review this one and I had discussed with him over slack how to solve that issue.

To fix my mistake here it goes: New Relic has a constant alert about poor performance of They Vote For You. @henare and I investigated the problem separately and we both came up with the same conclusion. The problem was around the MP's page. After looking at the code, I discussed with @henare that the best solution would be apply some sort of filters. So then he suggested to use the filters we already have in some other pages. Which I promptly agree and hence I created this PR.

Some screen shots now:

Now the URL: http://localhost:3000/people/representatives/cowper/luke_hartsuyker/divisions will load only the divisions of the last year in the DB, very similar to the divisions page http://localhost:3000/divisions

image
_Filter on the MP's page_

You can, of course, select another year: http://localhost:3000/people/representatives/cowper/luke_hartsuyker/divisions/2016

image

The feedback I want to have now is: Is the pagination implemented correctly? Is it ruby-ish enough? I changed a bit how the URL works. Although it does not break any previous links, the behaviour changes from loading all divisions to load divisions that happened on the last year in the DB. Is that okay? Is it gonna break any automated tool that reads They Vote For You? 

Regarding performance, I saw an huge improvement with pagination. It decreases the page from 30 seconds (in my local machine) to one or two at most.

spec/fixtures/static_pages/mp.php?mpn=Christine_Milne&mpc=Tasmania&house=senate&display=everyvote.html
mario-areias: Okay, I think I figured out why. I believe I fixed a bug that raised another one. In the `_divisions.html.haml` there is a bit of code like this: `%ol.divisions-list.list-unstyled{class: active_house_for_list_class(@house)}`. The method `active_house_for_list_class` is defined in `divisions_helper.rb`. The method is as follows:

```ruby
def active_house_for_list_class(house)
    if house == "representatives"
      "display-house-representatives"
    elsif house == "senate"
      "display-house-senate"
    elsif house == nil
      "display-house-all"
    end
end
```

The bug was the `@house` variable was always `nil`. And therefore only the class `display-house-all` was displayed. But I accidentally fixed that, which changed the HTML which I think is ok. However, there are no CSS classes defined for `display-house-senate` and `display-house-representatives`. There is only a class for `display-house-all` defined at `_division-list.scss`.  

After all explanation, my questions is: should I create the missing CSS classes with the same style as  `display-house-all`?

mario-areias:
I just added the date of the votes in the MPs page as well. Just like we do with divisions

image

equivalentideas:
> Is the pagination implemented correctly? Is it ruby-ish enough?

Looks good to me. I think the approach of mirroring the divisions index is great. That's the consistency that makes something really usable 💯 

> Although it does not break any previous links

❤️ ❤️ ❤️ 

> Is it gonna break any automated tool that reads They Vote For You?

I don't think anyone is scraping They Vote For You like this. If they were I'd hope they would help us add functionality to [the API](https://theyvoteforyou.org.au/help/data#divisions) instead ;-)

> I like the idea to have as few features as possible. To make sure we're focusing in high value, user oriented features.

🙌  Love this.

I'm going to make a couple of tiny changes for style, and one markup thing, and merge this.

equivalentideas:
I added a few bits on the end to hopefully make all the craziness more usable to people reading the site, and the code :) https://github.com/openaustralia/publicwhip/compare/a9848656a207...ae558edd17e7

Merged and 🚢 ed https://theyvoteforyou.org.au/people/representatives/grayndler/anthony_albanese/divisions

Not only is this a _huge_ improvement for users visiting these pages, but the refactoring is making this all easier to understand. Great contribution @mario-areias 🏆

Issue #1078, opened 2015-12-09
Member header layouts differ between pages where they shouldn't


equivalentideas:
Toggle between https://theyvoteforyou.org.au/people/representatives/wentworth/malcolm_turnbull and https://theyvoteforyou.org.au/people/representatives/wentworth/malcolm_turnbull/divisions to see that the space to the right of the members photo is wider on the member show page than the divisions index with member page:

![screen shot 2015-12-09 at 11 00 03 am](https://cloud.githubusercontent.com/assets/1239550/11672494/2d2d7fb8-9e65-11e5-8580-37d400fc38f5.png)
![screen shot 2015-12-09 at 10 59 53 am](https://cloud.githubusercontent.com/assets/1239550/11672495/2d305c24-9e65-11e5-9299-f4607ccc69fa.png)

PR #1134, opened 2017-03-28
Member header layouts made same for pages

Anishka0107:
This pull request aims to fix the issue https://github.com/openaustralia/publicwhip/issues/1078
Toggling between the member show and the divisions page, it is seen that the space to the right of the members photo is wider on the member show page than the divisions index with member page.
The divisions page has a difference of 10px whereas the member show page has a difference of 15px. 
To maintain consistency and considering that a 10px difference is more appropriate, the 15px difference on the member show page has been changed to 10px.

### DIFF
--- a/app/assets/stylesheets/members/_member-show.scss
+++ b/app/assets/stylesheets/members/_member-show.scss
@@ -26,7 +26,7 @@
 
   .media .pull-left {
     padding: 0;
-    margin-right: 15px;
+    margin-right: 10px;
   }
 
   .object-data-rebellion small,


henare:
This touches CSS so imma let @equivalentideas review this :smile:

equivalentideas:
Another interesting CSS bug! Today's my day.

Just to be clear, what we're talking about on this issue is the difference between the spacing on the right hand side of the Member’s photo on the [`members#show` view](https://theyvoteforyou.org.au/people/senate/wa/scott_ludlam) and the [`divisions#index` view (when it's for a single member’s divisions)](https://theyvoteforyou.org.au/people/senate/wa/scott_ludlam/divisions).

screen shot 2017-04-13 at 4 00 02 pm copy
screen shot 2017-04-13 at 4 00 10 pm copy

Poking around I also found a difference on the [`policies#show` with member view](https://theyvoteforyou.org.au/people/senate/wa/scott_ludlam/policies/91). The `member#show` view has the 15px and the other two have 10px on the right of the photo.

Looking at it, I also noticed that the icon for speeches was missing from the `divisions#index` version.

screen shot 2017-04-13 at 4 00 10 pm
screen shot 2017-04-13 at 4 00 02 pm

Which made me think “there's a style here that's meant to be applied but it's not for some reason".

So I went and looked up where that comes from, it's this block of styles under [`.member .page-header`](https://github.com/openaustralia/publicwhip/blob/master/app/assets/stylesheets/members/_member-show.scss#L20-L44). And guess what else is in there? The [`margin-right: 15px;`](https://github.com/openaustralia/publicwhip/blob/master/app/assets/stylesheets/members/_member-show.scss#L27-L30) that sets the spacing off the side of that photo!

So it's clear that we specifically overrode bootstrap here to increase the spacing from 10 to 15px for starters.

But what about the icon?

So what's happening here is that the block of styles that's needed to make the html in the partial `members/member_page_header` look right is all nested under `.members .page-header`. The member class is coming from the [`body` element and is set by which controller has rendered the page](https://github.com/openaustralia/publicwhip/blob/master/app/views/layouts/application.html.haml#L26). So when the partial [`members/member_page_header` is rendered from the DivisionsController](https://github.com/openaustralia/publicwhip/blob/master/app/views/divisions/index_with_member.html.haml#L6) it doesn't have the `.members` class and none of the styles apply.

So that's why the spacing is different and why the icon is missing.

Having the controller interact with the css in this way isn't good practice (probably my bad!)—but that's how it works here.

But we can get away from the by making this block of css less brittle, and not dependent on the controller.

I think a good solution here is to change the selector `.members .page-header` to `.member-page-header` in `members/_member-show.scss:` line 20, and add that class to the [`members/member_page_header` partial](https://github.com/openaustralia/publicwhip/blob/master/app/views/members/_member_page_header.html.haml#L1) so it has `.member-page-header` and `.page-header`. That way it'll get the bootstrap stuff from `.page-header` and all our overrides from our custom scss.

Then we can add our new `.member-page-header` class to the [policy with member view](https://theyvoteforyou.org.au/people/senate/wa/scott_ludlam/policies/91) so it gets the proper spacing too :)

Then we have more robust css thats isolate from the controller code, and all the styles will be consistent.

Sorry to go into such detail again! I just am getting excited about making the css better :) Does that all make sense and also sound like a good idea?

Anishka0107:
@equivalentideas This seems pretty interesting and I didn't realize that there was so much to it. I would like to work on it!

equivalentideas:
> This seems pretty interesting and I didn't realize that there was so much to it. I would like to work on it!

Glad its interesting :D Do you feel comfortable updating this PR to make that change?

equivalentideas:

PR #1135, opened 2017-03-28
Recent Changes page header layout changed

### LINKED ISSUE #989: Extra separating line on Recent Edits page seems unnecessary
I'm not sure it's mean to be there.

selection_001

Anishka0107:
Fixes #989.

The Recent Changes page https://theyvoteforyou.org.au/history contained two lines below the header, the upper one coming from it as it was a page-header, and the lower one coming from the list of recent changes.
The upper line looked unnecessary, so it was removed. 
Also, to make the design consistent with the other pages layout, the header layout was changed and made similar to other pages like https://theyvoteforyou.org.au/help/faq.
In this commit, the Recent Changes page does not contain the unnecessary line and also has a consistent (similar to other pages) header layout.

This is how the Recent Changes page now looks like (in the commit).
testing

### DIFF
--- a/app/views/home/history.html.haml
+++ b/app/views/home/history.html.haml
@@ -1,6 +1,8 @@
 - content_for :title, 'Recent edits'
 - set_meta_tags description: "Recent edits to divisions and policies by the #{Settings.project_name} community."
-.page-header
-  %h1 All edits made in the last week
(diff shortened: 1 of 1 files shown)


henare:
Deployed :rocket:

PR #1136, opened 2017-03-28
Adding Generic 404 page error

essymo:
adding the line;

config.exceptions_app = self.routes in (config/application.rb) to tell the application's router to handle the errors. If the URL is missing or misspelled then it should redirect to the 404 page.


equivalentideas:
Looks good @essymo I'll merge and ship this :shipit: 

Thanks for this contribution!

In the future, I think you could be more descriptive in your commit message. e.g. what you had in your PR description would be useful information for the commit. It's good that from your message people can understand the intention of the change to the line though 👍

PR #1138, opened 2017-03-29
Remove padding from introductory paragraph on MP votes on policy page

### LINKED ISSUE #1095: Vertical alignment on *MP votes on policy* page
The introductory paragraph is a bit out of whack. See screenshot below.

before

I haven't found the problem anywhere else.

Anishka0107:
This commit aims to fix the issue #1095
The MP votes on policy page has a problem with the vertical alignment as mentioned in the issue. The introductory paragraph does not align vertically with the other part of the page, it is shifted to the right because of the use of col-sm-7 class of Bootstrap, which introduces a padding of 15px.
To override this, a new id is added to this paragraph with the left padding made 0, to shift it to the right position.

Now, it looks like
fixedupstuff

Closes #1095

### DIFF
--- a/app/assets/stylesheets/policies/_policy-voter-comparision.scss
+++ b/app/assets/stylesheets/policies/_policy-voter-comparision.scss
@@ -130,3 +130,7 @@
 *:hover > .anchor {
   display: block;
 }
+
+#col-sm-fixup {
+  padding-left: 0;
+}
--- a/app/assets/stylesheets/policies/_policy-voter-comparision.scss
+++ b/app/assets/stylesheets/policies/_policy-voter-comparision.scss
@@ -131,6 +131,6 @@
   display: block;
 }
 
-#col-sm-fixup {
+#voting-comparision {
   padding-left: 0;
 }
--- a/app/views/policies/show_with_member.html.haml
+++ b/app/views/policies/show_with_member.html.haml
@@ -16,7 +16,7 @@
 
 = render "social_share"
 
-%p#col-sm-fixup.col-sm-7
+%p#voting-comparision.col-sm-7
   How
   = link_to @member.name, @member
   voted compared to someone who believes that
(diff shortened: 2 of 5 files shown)


Anishka0107:
@equivalentideas The Travis build fails. Though I have tested it on my local machine I don't know why it's failing.
Could you help.

mario-areias:
Hi @Anishka0107 ,

This is the same issue I described [here](https://github.com/openaustralia/publicwhip/pull/1139#issuecomment-290640515). Basically, you changed the UI and now the tests failed. Have a look and see if the changes in these tests are related to what you did. If they are, just push the modified files and you should be alright. If they are not, then you need to investigate and understand why that occurred.

Anishka0107:
Yes, @mario-areias They are related to the changes I made. Thanks a lot for the help. Is there anything I need to do next?

mario-areias:
Just commit the changed files. It will make the tests green 😄

app/assets/stylesheets/policies/_policy-voter-comparision.scss
```diff
@@ -129,4 +129,8 @@
 
 *:hover > .anchor {
   display: block;
-}
\ No newline at end of file
+}
+
+#col-sm-fixup {
```
henare: @Anishka0107 I think we need a different name than "fixup". Since we're always fixing things up, if we did this then pretty soon everything would be called "fixup"! :smile: 

Since it's an ID, perhaps describe what that element is?

Anishka0107:
@henare
I have changed the name of the ID to voting-comparision. Does that seem fine?

henare:
@Anishka0107 looks good to me. I'm going to request a review from @equivalentideas because he's very opinionated about our CSS :smile:

equivalentideas:

Issue #726, opened 2014-10-13
Present tense used for rebellions about former members


henare:
http://publicwhip-rails.openaustraliafoundation.org.au/members/representatives/o'connor/tony_crook

> Rebels on 7.7% of votes.

then

> Attended 87% of votes (440 of 504).

It should be

> Rebelled ..

mlandauer:
I think the PR above fixed this

PR #1139, opened 2017-03-30
Use present tense for former member rebellions

Anishka0107:
This pull request aims to fix the issue https://github.com/openaustralia/publicwhip/issues/726
Initially, on the Member Show page, 'Rebels' was used for current and former members as well. For former members, it should be 'Rebelled' instead.
After this commit, 'Rebels' would be used for current members and 'Rebelled' for former members.

Now, it looks like: 
newupload

### DIFF
--- a/app/views/members/_member_page_header.html.haml
+++ b/app/views/members/_member_page_header.html.haml
@@ -13,17 +13,30 @@
           %p.member-data.object-data
             - if member.person.rebellions_fraction
               %span.member-rebellions.object-data-rebellion
-                -if member.person.rebellions_fraction == 0
-                  Never rebels
-                -else
-                  -# TODO: Should this be an absolute count rather than percentage?
-                  -# Maybe it's good to show it as a percentage because it highlights rarity?
-                  Rebels
-                  = fraction_to_percentage_display(member.person.rebellions_fraction)
-                  of the time
-                  -# TODO: add helper tooltip for rebellions
-                  -# link_to "explain...", help_faq_path(anchor: "clarify")
+                - if !member.currently_in_parliament?
+                  - if member.person.rebellions_fraction == 0
+                    Never rebelled
+                  - else
+                    -# TODO: Should this be an absolute count rather than percentage?
+                    -# Maybe it's good to show it as a percentage because it highlights rarity?
+                    Rebelled
+                    = fraction_to_percentage_display(member.person.rebellions_fraction)
+                    of the time
+                    -# TODO: add helper tooltip for rebellions
+                    -# link_to "explain...", help_faq_path(anchor: "clarify")
+                - else
+                  - if member.person.rebellions_fraction == 0
+                    Never rebels
+                  - else
+                    -# TODO: Should this be an absolute count rather than percentage?
+                    -# Maybe it's good to show it as a percentage because it highlights rarity?
+                    Rebels
+                    = fraction_to_percentage_display(member.person.rebellions_fraction)
+                    of the time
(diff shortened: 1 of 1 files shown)


Anishka0107:
@equivalentideas 
The tests fail again, both locally and on Travis. Looking at the job log, the error displayed is "Don't match. Writing over file in spec/fixtures/static_pages. Do a git diff." 
I think this is because of the function https://github.com/openaustralia/publicwhip/blob/9374e4db627ae25a16d49adf0ae370840f05335f/spec/html_compare_helper.rb#L72.
It checks if the text on the website has changed in the commit, and this is what I have done!
Am I correct in recognising the cause of the failure sir?

app/views/members/_member_page_header.html.haml
```diff
@@ -13,17 +13,30 @@
...
-                  Never rebels
-                -else
-                  -# TODO: Should this be an absolute count rather than percentage?
-                  -# Maybe it's good to show it as a percentage because it highlights rarity?
-                  Rebels
-                  = fraction_to_percentage_display(member.person.rebellions_fraction)
-                  of the time
-                  -# TODO: add helper tooltip for rebellions
-                  -# link_to "explain...", help_faq_path(anchor: "clarify")
+                - if !member.currently_in_parliament?
```
mario-areias: Hi @Anishka0107 ,

Thanks for this PR. I would like to suggest a couple of suggestions. I noticed you duplicated the `if` block. That's usually not very good, because if we need to change the logic of this block, now we will need to do in two places. There is a principle for that if you are interested to read more about it. It's called [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

Ruby also can help here. Instead to duplicate the `if` block, you can just added two ifs to add the right messages. For example, you can replace the string `Rebels` for the code below:

```ruby
= "Rebels" if member.currently_in_parliament?
= "Rebelled" if !member.currently_in_parliament?
```
This will print `Rebels` when the `member.currently_in_parliament` and `Rebelled` when the member is not. So then you don't need to duplicate the `if` block :smile:

mario-areias:
@Anishka0107 the tests fails, because you have changed the HTML page structure. These tests are not very good and lots of people get confused by it. 

Anyway, just check if the changes in these files are related to your commits. For example, it might have changed some `Rebels` to `Rebelled`. If that's the case, this is expected (that's what your commit was about). If there are other changes not related to your code, it might be a unintended collateral effect of your commit. So you need to investigate to understand why that's happening. 

Does that make sense?

Anishka0107:
@mario-areias I have made the changes as you suggested. Could you please take a look?

app/views/members/_member_page_header.html.haml
Anishka0107: @mario-areias I have made the changes as you suggested. Could you please take a look?

mario-areias:
Nice. Now you can add the files changed by the tests, so it makes the tests green. Try to add a meaningful commit message as well. Describing what you did. For example, for your previous commit the message could be something like "Removed duplicated logic and replaced with two ifs for members page". Or something like that, so then when we look at the git history we know what you wanted to accomplish 😄

Anishka0107:
@mario-areias I will be careful with the commit message texts next time. However, I do not know how to add files changed by the tests. Could you help me with that?

Anishka0107:
Closing this PR, created a new clean PR. Thanks a lot @mario-areias for the help
Will take into account the fact that we should not close and open a new PR for commits on the same issue from the next time.

PR #1141, opened 2017-03-31
Use past tense for former member rebellions

Anishka0107:
This commit aims to fix the issue openaustralia#726
Initially, on the Member Show page, 'Rebels' was used for current and former members as well. For former members, it should be 'Rebelled' instead.
After this commit, 'Rebels' would be used for current members and 'Rebelled' for former members.

For former members, it looks like :
newupload

### DIFF
--- a/app/views/members/_member_page_header.html.haml
+++ b/app/views/members/_member_page_header.html.haml
@@ -14,11 +14,13 @@
             - if member.person.rebellions_fraction
               %span.member-rebellions.object-data-rebellion
                 -if member.person.rebellions_fraction == 0
-                  Never rebels
+                  = "Never rebels" if member.currently_in_parliament?
+                  = "Never rebelled" if !member.currently_in_parliament?
                 -else
                   -# TODO: Should this be an absolute count rather than percentage?
                   -# Maybe it's good to show it as a percentage because it highlights rarity?
-                  Rebels
+                  = "Rebels" if member.currently_in_parliament?
+                  = "Rebelled" if !member.currently_in_parliament?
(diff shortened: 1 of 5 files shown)


mario-areias:
Looks good to me @Anishka0107 !

Well done 😄

app/views/members/_member_page_header.html.haml
```diff
@@ -14,11 +14,13 @@
             - if member.person.rebellions_fraction
               %span.member-rebellions.object-data-rebellion
                 -if member.person.rebellions_fraction == 0
-                  Never rebels
+                  = "Never rebels" if member.currently_in_parliament?
+                  = "Never rebelled" if !member.currently_in_parliament?
```
equivalentideas: A simpler way to implement this would be to use the [ternary operator](https://en.wikipedia.org/wiki/%3F:#Ruby) in ruby (e.g. a one line if/else statement) 

```
- member.currently_in_parliament? ? "Never rebels" : "Never rebelled"
```

That would also be a more expressive way to capture this logic.

equivalentideas:
Thanks for reviewing this as well @mario-areias 👍 

Also, @Anishka0107 good work with the screenshot and descriptive PR and commit message. It makes it really easy to review ❤️

app/views/members/_member_page_header.html.haml
Anishka0107: I have added another commit @equivalentideas with the changes you suggested

equivalentideas:
Great work @Anishka0107 🚀 I've merged this and will deploy!

Issue #1076, opened 2015-12-01
When you log out from a page that requires you to be logged in, you get taken to the sign in page and told to sign back in


equivalentideas:
This is a small corner case at this point, but if you log out from a page that requires you to be logged in, you get taken back to that page, which sends you back to log in again. Pretty confusing experience.

essymo:
I think its because we have set the session to previous url even when users log out. I think its better if we redirect logged out users to home page. What do you think? :)

equivalentideas:
> This is a small corner case at this point, but if you log out from a page that requires you to be logged in, you get taken back to that page, which sends you back to log in again. Pretty confusing experience.

I think it would be good to make a change only for this case. If you're on a page that doesn't require you to be logged in, and you log out, it could be because you want to see what they page is like when you're logged out. Being redirected to the home page in that case would be quite frustrating.

Are there any examples of how other projects deal with this? Like Facebook, Wikipedia or GitHub for example?

essymo:
oh yeaah I see what you mean being redirected to the home page is not a good option :). I guess I did not see it in that perspective that a user may log out to see what the page looks like,most of the time people log out when they are done doing what they were doing..well I do that :)  I think this is the code in question..
` def store_location
    # store last url as long as it isn't a /users path
    session[:previous_url]= request.fullpath unless request.fullpath =~ /\/users/
  end

  def after_sign_in_path_for(resource)
    path = stored_location_for(resource) || session[:previous_url]
    path.nil? || path == root_path ? user_welcome_path : path
  end

  def after_sign_out_path_for(resource)
    request.referer ? URI.parse(request.referer).path : root_path
  end
end
`
 I am not sure if this is a bug because for example, when you are on a page that does not require you to be logged in, lets say division page,if you log out you will be redirected to the same division page which in this case is the ( previous url) awesome right?! :) the same thing happens with edit page but since you have to be logged in, it sends you to the log in for you to be allowed to access the edit page. If its a bug maybe shed some more light on this :) Any ideas?

I reproduced this on Facebook and Github,so what happens is that when you log out from a page that requires you to be logged in you are taken to the sign_up and sign_in page. When you log back you are taken to the home page.

equivalentideas:
>  If its a bug maybe shed some more light on this :)

I think after you've spelt it out like this, and looked at other examples, I don't think it is a bug any more :) Nice one @essymo I'll close this issue!

PR #1142, opened 2017-04-02
When users log out redirect them to the home page

essymo:
This pull request aims to fix issue #1076.
Currently when a user logs out you get taken to the sign in page and told to sign back in and then you  get taken back to that same page.I think this is because we have set the session to previous URL. I suggest that once users click log out they should be redirected to the home page.


equivalentideas:
@essymo worked out that this isn't a bug #1076 so closing this PR

PR #1143, opened 2017-04-02
Made title of Divisions history edit page appropriate

Anishka0107:
This commit aims to fix the issue openaustralia#798
The date and title have now been sent to new lines with slightly different layouts to distinguish between them and provide an easy to distinguish title.

This is what it looks like now:
newtoup

### DIFF
--- a/app/views/divisions/history.html.haml
+++ b/app/views/divisions/history.html.haml
@@ -2,14 +2,16 @@
   = Diffy::CSS
 
 - content_for :title do
-  Division Description Changes —
-  = @division.original_name
-  —
-  = formatted_date(@division.date)
+  Division Description Changes
 - set_meta_tags description: "All changes made to the description and title of the division: #{@division.name}"
 
 .page-header
-  %h1= yield :title
+  %h1
+    = yield :title
+  %h2
+    = @division.original_name
+    %small.pre-title
+      = formatted_date(@division.date)
 
 %p
   All changes made to the description and title of this
@@ -41,3 +43,4 @@
           = Diffy::Diff.new(wiki_motion.previous_title.strip, wiki_motion.title.strip).to_s(:html).html_safe
           %p Description
           = Diffy::Diff.new(wiki_motion.previous_description, wiki_motion.description).to_s(:html).html_safe
+
--- a/app/views/divisions/history.html.haml
+++ b/app/views/divisions/history.html.haml
@@ -1,17 +1,16 @@
 %style
   = Diffy::CSS
 
-- content_for :title do
-  Division Description Changes
+- content_for :title, "Division Description Changes"
 - set_meta_tags description: "All changes made to the description and title of the division: #{@division.name}"
 
 .page-header
+  %h1= yield :title
   %h1
-    = yield :title
-  %h2
-    = @division.original_name
-    %small.pre-title
-      = formatted_date(@division.date)
+    .division-title.long-title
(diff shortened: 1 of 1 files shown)


app/views/divisions/history.html.haml
```diff
@@ -2,14 +2,16 @@
   = Diffy::CSS
 
 - content_for :title do
-  Division Description Changes —
-  = @division.original_name
-  —
-  = formatted_date(@division.date)
+  Division Description Changes
```
henare: I don't think this needs to be nested in a `do` block any more so you can inline it into the `content_for` method call.

app/views/divisions/history.html.haml
```diff
@@ -2,14 +2,16 @@
...
 - set_meta_tags description: "All changes made to the description and title of the division: #{@division.name}"
 
 .page-header
-  %h1= yield :title
+  %h1
+    = yield :title
```
henare: Any reason this was split onto its own line? It's more concise how it was before.

app/views/divisions/history.html.haml
```diff
@@ -2,14 +2,16 @@
...
+  %h2
+    = @division.original_name
+    %small.pre-title
```
henare: Should the date be nested in the h2? I'm really not sure about this - we should check and make it consistent with the rest of the app (I'm guessing we use this pattern elsewhere?).

Anishka0107:
@henare I have included the changes you suggested and used h1 as well. I tried making it consistent with the page https://theyvoteforyou.org.au/divisions/senate/2017-03-30/3

app/views/divisions/history.html.haml
```diff
@@ -1,15 +1,16 @@
...
-  Division Description Changes —
-  = @division.original_name
-  —
-  = formatted_date(@division.date)
+- content_for :title, "Division Description Changes"
 - set_meta_tags description: "All changes made to the description and title of the division: #{@division.name}"
 
 .page-header
   %h1= yield :title
+  %h1
```
Anishka0107: @henare I have used another h1 to introduce a new line between the two titles as it looks better with the space in between these two. Is that okay?

henare: Thanks for flagging this @Anishka0107. I'm not sure about this, @equivalentideas what do you think?

equivalentideas:
> I have used another h1 to introduce a new line between the two titles as it looks better with the space in between these two. Is that okay?

@Anishka0107 I think we should use a h1 for the 'division edit history' (or some wording that's more consistent with our 'history' wording https://theyvoteforyou.org.au/policies/50/history https://theyvoteforyou.org.au/history ), and then a h2 for the division title. But still use all the classes and markup pattern for the division title from the division show page.

That could come out looking like this.

screen shot 2017-05-16 at 4 55 48 pm

I think it's particularly important that the division title matches the style of the show page here, unless there's some strong reason to break the pattern.

Does that help clear up the next steps for this @Anishka0107 ? Thanks for all your contributions as always :)

Anishka0107:
Hi @equivalentideas, I really wish to work on this and the other PRs too. Thanks for keeping them open for so long. I would resume work on them after a few weeks. Is it fine to leave them open for the time being?

equivalentideas:
@Anishka0107 yep no worries! Speak soon :)

equivalentideas:

PR #1145, opened 2017-04-04
Extracting logic from view and adding it to helpers

Anishka0107:
The logic for member rebellion (whether it should be Rebelled/Rebels/Never rebels/Never rebelled) in the member page header file was complicated and hence was moved to view helpers to make it look simple and consistent.

### DIFF
--- a/app/helpers/members_helper.rb
+++ b/app/helpers/members_helper.rb
@@ -59,14 +59,13 @@ module MembersHelper
     text.html_safe + "."
   end
 
-  def member_rebellion(member)
+  def member_rebellion_record_sentence(member)
     if member.person.rebellions_fraction == 0
-      text = member.currently_in_parliament? ? "Never rebels" : "Never rebelled"
+      member.currently_in_parliament? ? "Never rebels" : "Never rebelled"
     else
       # TODO: Should this be an absolute count rather than percentage?
       # Maybe it's good to show it as a percentage because it highlights rarity?
-      text = (member.currently_in_parliament? ? "Rebels " : "Rebelled ") + fraction_to_percentage_display(member.person.rebellions_fraction) + " of the time"
+      (member.currently_in_parliament? ? "Rebels " : "Rebelled ") + fraction_to_percentage_display(member.person.rebellions_fraction) + " of the time"
     end
-    text
   end
 end
--- a/app/helpers/members_helper.rb
+++ b/app/helpers/members_helper.rb
@@ -58,4 +58,15 @@ module MembersHelper
     end.to_sentence
     text.html_safe + "."
   end
+
+  def member_rebellion(member)
+    if member.person.rebellions_fraction == 0
+      text = member.currently_in_parliament? ? "Never rebels" : "Never rebelled"
+    else
+      # TODO: Should this be an absolute count rather than percentage?
+      # Maybe it's good to show it as a percentage because it highlights rarity?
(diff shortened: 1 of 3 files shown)


app/views/members/_member_page_header.html.haml
```diff
@@ -13,16 +13,9 @@
...
-                  = member.currently_in_parliament? ? "Never rebels" : "Never rebelled"
-                -else
-                  -# TODO: Should this be an absolute count rather than percentage?
-                  -# Maybe it's good to show it as a percentage because it highlights rarity?
-                  = member.currently_in_parliament? ? "Rebels" : "Rebelled"
-                  = fraction_to_percentage_display(member.person.rebellions_fraction)
-                  of the time
-                  -# TODO: add helper tooltip for rebellions
-                  -# link_to "explain...", help_faq_path(anchor: "clarify")
+                #{member_rebellion(member)}
```
equivalentideas: You can use the `=` syntax here to be consistent with the rest of the file:

```
= member_rebellion(member)
```

app/helpers/members_helper.rb
```diff
@@ -58,4 +58,15 @@ def member_history_sentence(member)
     end.to_sentence
     text.html_safe + "."
   end
+
+  def member_rebellion(member)
+    if member.person.rebellions_fraction == 0
+      text = member.currently_in_parliament? ? "Never rebels" : "Never rebelled"
+    else
+      # TODO: Should this be an absolute count rather than percentage?
```
equivalentideas: Good work to keep the comment here 👍

app/helpers/members_helper.rb
```diff
@@ -58,4 +58,15 @@ def member_history_sentence(member)
     end.to_sentence
     text.html_safe + "."
   end
+
+  def member_rebellion(member)
```
equivalentideas: I think this method could have a more descriptive name, like `member_rebellion_record_sentence` or something like that, so you don't have to look at the code to know what to expect :)

app/helpers/members_helper.rb
```diff
@@ -58,4 +58,15 @@ def member_history_sentence(member)
...
+    if member.person.rebellions_fraction == 0
+      text = member.currently_in_parliament? ? "Never rebels" : "Never rebelled"
+    else
+      # TODO: Should this be an absolute count rather than percentage?
+      # Maybe it's good to show it as a percentage because it highlights rarity?
+      text = (member.currently_in_parliament? ? "Rebels " : "Rebelled ") + fraction_to_percentage_display(member.person.rebellions_fraction) + " of the time"
+    end
+    text
```
equivalentideas: Do you need this `text` to return the value? I think if you remove that and the assignment above it will just return the values you want.

equivalentideas:
Really nice work @Anishka0107 🎆 I've merged this. This is a really nice little clean up that will make this app easier to develop as we go 🔧 Thanks for your contribution!

Nice use of mocking to keep the tests isolated 👌 

I made a few changes to make this a little easier to read and to be consistent with the rest of the app in terms of hash style and whitespace/empty lines.

You can walk through those commits here https://github.com/openaustralia/publicwhip/compare/b08bd60bdb1c...ef80176b597f

[One change I made](https://github.com/openaustralia/publicwhip/commit/248a46c0a75bdb6d6d1837056a8143f2e08b445c?diff=split) was to make it so each `context` block only changed one variable we care about. That also lets use extract the shared test setup :)

Great work 👍

Anishka0107:
Thanks for the review @equivalentideas :)
I will take into account these changes from the next time and will try simplifying things as much as possible.

PR #1147, opened 2017-04-09
Performance of division page

mario-areias:
This PR intend to improve the performance of the display of a single division page (https://theyvoteforyou.org.au/divisions/senate/2017-02-07/1).

Before this PR this page was doing around 4 queries for each line of the following table 
image

My approach was to try to load more information in few queries to avoid doing unnecessary round trips to the DB. In my local machine, I got an improvement of 100-150 ms in average, which means 20% of the initial 700 ms. Of course, New Relic will tell us if there is actually an improvement on that area and how much that's it.

I also added the `bullet` gem to identify possible problems in our queries, although I didn't find quite useful, maybe others can use it better.

I do not have the context around the code involved to render the division page and I am pretty sure more refactoring is need to improve the changes I've done. So please feel free to suggest editions.

I also created a couple of tests for the `DivisionsController#show` method which are not very good yet, but it's a start. Again, feel free to suggest improvement here as well.

There should not be any difference of behaviour for the final user, except to a slightly improvement on the performance.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -96,6 +96,7 @@ group :development do
   gem 'mina'
   gem 'mina-multistage', require: false
   gem 'brakeman', require: false
+end
 
 group :test, :development do
   gem 'rspec-rails', '~> 3.5'
--- a/Gemfile
+++ b/Gemfile
@@ -95,9 +95,7 @@ group :development do
 
   gem 'mina'
   gem 'mina-multistage', require: false
-
   gem 'brakeman', require: false
-end
 
 group :test, :development do
   gem 'rspec-rails', '~> 3.5'
--- a/app/controllers/divisions_controller.rb
+++ b/app/controllers/divisions_controller.rb
@@ -238,6 +238,6 @@ class DivisionsController < ApplicationController
     Division.in_house(house)
     .joins(:division_info, :whips)
     .includes(:division_info, :whips)
-    .where(date: date, number: number).first
+    .find_by(date: date, number: number)
   end
 end
--- a/app/controllers/divisions_controller.rb
+++ b/app/controllers/divisions_controller.rb
@@ -109,30 +109,37 @@ class DivisionsController < ApplicationController
 
   def show
     house = params[:house]
+    date = params[:date]
+    number = params[:number]
 
     @division = Division.in_house(house)
     @division = @division.joins(:division_info, :whips)
     @division = @division.includes(:division_info, :whips)
-    @division = @division.where!(date: params[:date], number: params[:number]).first!
+    @division = @division.where(date: date, number: number).first
 
-    @rebellions = @division.votes.rebellious.order("members.last_name", "members.first_name") if @division.rebellions > 0
-    @whips = @division.whips.order(:party)
-    @votes = @division.votes.joins(:member).includes(:member).order("members.party", "vote", "members.last_name", "members.first_name")
+    if @division.nil?
+      render 'home/error_404', status: 404 unless @division
+    else
+      @rebellions = @division.votes.rebellious.order("members.last_name", "members.first_name") if @division.rebellions > 0
+      @whips = @division.whips.order(:party)
+      @votes = @division.votes.joins(:member).includes(:member).order("members.party", "vote", "members.last_name", "members.first_name")
 
-    # If a member is included
-    if params[:mpn] && params[:mpc]
-      name = params[:mpn].gsub("_", " ")
-      electorate = params[:mpc].gsub("_", " ")
-      # TODO Also ensure that the member is current on the date of this division
-      member = Member.in_house(house).with_name(name).
+      # If a member is included
+      if params[:mpn] && params[:mpc]
+        name = params[:mpn].gsub("_", " ")
+        electorate = params[:mpc].gsub("_", " ")
+        # TODO Also ensure that the member is current on the date of this division
+        member = Member.in_house(house).with_name(name).
         where(constituency: electorate).first
-      @member = member.person.member_who_voted_on_division(@division)
-    end
-    @members = Member.in_house(house).current_on(@division.date).
+        @member = member.person.member_who_voted_on_division(@division)
+      end
+      @members = Member.in_house(house).current_on(@division.date).
       joins("LEFT OUTER JOIN votes ON members.id = votes.member_id AND votes.division_id = #{@division.id}").
       order("members.party", "vote", "members.last_name", "members.first_name")
 
-    @members_vote_null = @members.where("votes.id IS NULL")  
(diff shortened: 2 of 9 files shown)


Gemfile
```diff
@@ -94,6 +94,7 @@ group :development do
 
   gem 'mina'
   gem 'mina-multistage', require: false
+  gem 'bullet'
```
henare: >I also added the bullet gem to identify possible problems in our queries, although I didn't find quite useful, maybe others can use it better.

If we're not using this I don't think it should be added.

app/controllers/divisions_controller.rb
```diff
@@ -219,4 +233,11 @@ def get_member
           .where(constituency: electorate)
           .order(entered_house: :desc).first
   end
+
+  def get_division(house, date, number)
```
henare: This smells like it should be a model class method or possibly scope?

app/models/vote.rb
```diff
@@ -29,8 +29,12 @@ def self.noes
     where("votes.vote = 'no'")
   end
 
-  def rebellion?
-    !free? && vote != whip_guess
+  def rebellion?(whip=nil)
```
henare: Style-only: [spaces around the `=` are preferred](https://github.com/bbatsov/ruby-style-guide#spaces-around-equals).

app/controllers/divisions_controller.rb
```diff
@@ -219,4 +233,11 @@ def get_member
...
+    Division.in_house(house)
+    .joins(:division_info, :whips)
+    .includes(:division_info, :whips)
+    .where(date: date, number: number).first
```
henare: Use `#find_by` instead of `#where` and `#first` here.

Gemfile
mario-areias: I think it would be useful if you and @equivalentideas can have a look and test it. It might be useful for you guys. If you don't find useful I can delete it, but I think it's worth to have you both testing it before that.

app/controllers/divisions_controller.rb
mario-areias: Do you think so? I thought this method wasry ve particular to this controller. I extracted as a method to leave the code a bit more readable, but otherwise it would live inside the `show` method. I would rather move if we have repetition. However, I have no strong opinions about it, so if you want me to move, I can move it :smile:

app/models/vote.rb
mario-areias: Sure. But what do you think about this method overall? I did that to improve the performance, but I hated the end resutlts. I don't have complete understanding about what this method does, so I'm not sure how to improve it. If you have any suggestions, that would be awesome!

app/views/divisions/_summary.html.haml
```diff
@@ -57,8 +57,8 @@
                   %span.sr-only show members
 
           -# Show members in this party that are present in this division
-          - division.votes.joins(:member).where(members: {party: whip.party}).includes(:member).order("members.party", "vote", "members.last_name", "members.first_name").each do |vote|
-            %tr.member-row{class: [('collapse party-member-row' unless whip.whipless? || whip.possible_votes == 1), ('rebel' if vote.rebellion?), ('member-row-' + whip.party.parameterize)]}
+          - @votes.select {|vote| vote.member.party == whip.party}.each do |vote|
+            %tr.member-row{class: [('collapse party-member-row' unless whip.whipless? || whip.possible_votes == 1), ('rebel' if vote.rebellion?(whip)), ('member-row-' + whip.party.parameterize)]}
```
henare: This looks really complex. Maybe we should extract this into a helper?

app/controllers/divisions_controller.rb
henare: If you're not fussed neither am I. We can always refactor later if the smell gets stronger :smile:

Gemfile
henare: It might be useful but I'd rather install it when we need it otherwise it's really easy to build up a whole bunch of "useful" gems that you never use :smile:

equivalentideas: > I also added the bullet gem to identify possible problems in our queries, although I didn't find quite useful, maybe others can use it better.

I think exclude that from this PR if you didn't find it useful. I've only heard about it, but never actually used it.

mario-areias: No problem,. I will delete it then.

app/models/vote.rb
```diff
@@ -29,8 +29,12 @@ def self.noes
...
   end
 
-  def rebellion?
-    !free? && vote != whip_guess
+  def rebellion?(whip = nil)
+    if whip.nil?
+      !free? && vote != whip_guess
+    else
+      !whip.free? && vote != whip.whip_guess
+    end
```
henare: @mario-areias wrote on outdated diff:

>Sure. But what do you think about this method overall? I did that to improve the performance, but I hated the end resutlts. I don't have complete understanding about what this method does, so I'm not sure how to improve it. If you have any suggestions, that would be awesome!

How does it improve performance?

I'm not really sure this is doing what you think it's doing. In the first version of the `if` statement it's using the [delegated methods](https://github.com/mario-areias/publicwhip/blob/fc010657f191f1d50601e18a4b84096f8ca0431f/app/models/vote.rb#L6) and the second version it's just chaining the methods onto the [`#whip` method](https://github.com/mario-areias/publicwhip/blob/fc010657f191f1d50601e18a4b84096f8ca0431f/app/models/vote.rb#L9-L11) directly. I can't see how there would be a difference between the two.

mario-areias: The first `if` it uses delegate (as you said). When it call this method it does a query to get the whip and then the other queries related to that. The second `if` it gets from the `whip` from the parameter  (which is already loaded and then avoiding one extra query). 

It seems silly, but when in a for loop, that can avoid several queries to the DB. But it's ugly and I don't like it. If you have another suggestion to avoid this other query that would be good :smile:

henare: Interesting, thanks @mario-areias!

Could we memoize `#whip` instead?

mario-areias: Hum, I will try that. But not sure if it will work. `memoize` works if you call the same method repeatedly, but I am not sure if that's the case here (I don't remember I opened this a few days ago :smiley_cat: ). IIRC the `whip` was loaded first and then `vote` was loaded later. That's why I did what I did. But as I said, I will try your suggestoin regardless :smile:

mario-areias:
@henare I ended up leaving the `rebellion` method in the `vote` class unchanged and moved the additional logic to the `divisions_helper`. The reason for that is I don't fully understand the relationships of `division`, `vote` and `whip` tables and therefore I preferred to extract that logic to a different file.

Let me know your thoughts on it.

PR #1149, opened 2017-04-11
Avoided multiple render methods to be called at DivisionsController

### LINKED ISSUE #1148: Malformed date in URL 500s instead of 404ing
e.g. https://theyvoteforyou.org.au/people/representatives/wentworth/malcolm_turnbull/divisions/2012-

`[They Vote For You/production] AbstractController::DoubleRenderError: Render and/or redirect were called multiple times in this action. Please note that you may only call render OR redirect, and at most once per action. Also note that neither redirect nor render terminate execution of the action, so if you want to exit an action after redirecting, you need to do something like "redirect_to(...) and return".`

## Backtrace

    line 28 of [PROJECT_ROOT]/app/controllers/divisions_controller.rb: index

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/37542/faults/33226459)

mario-areias:
Fixes #1148 

The issue was that the `render` method does stop prevent further actions to continue, so depending on the case the controller could call the render method twice. 

For example here: https://theyvoteforyou.org.au/people/representatives/wentworth/malcolm_turnbull/divisions/2012-

This should render a 404 page due to the fact the date is bad formatted. However, because it's also a `Member` page it also renders the `Member` page as well. This PR fixed that making sure, it renders the 404 page avoiding to go further at the code.

As a second thought, I think it would be nice to have a different method and different route for the member's division page.

### DIFF
--- a/app/controllers/divisions_controller.rb
+++ b/app/controllers/divisions_controller.rb
@@ -21,13 +21,16 @@ class DivisionsController < ApplicationController
         @date_start, @date_end, @date_range = get_date_range(params[:date])
       rescue ArgumentError
         render 'home/error_404', status: 404
+        return
       end
 
       if @member
         @divisions = @member.divisions_they_could_have_attended_between(@date_start, @date_end)
         render 'index_with_member'
+        return
       else
         render 'members/member_not_found', status: 404
+        return
       end
     else
       @sort = params[:sort]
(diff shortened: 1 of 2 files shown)


app/controllers/divisions_controller.rb
```diff
@@ -21,13 +21,16 @@ def index
         @date_start, @date_end, @date_range = get_date_range(params[:date])
       rescue ArgumentError
         render 'home/error_404', status: 404
+        return
```
henare: Can you just do this on the same line?

```ruby
return render 'home/error_404', status: 404
```

henare:
Nice work @mario-areias :ok_hand:

PR #1150, opened 2017-04-11
Added static analysis tool for security

mario-areias:


mario-areias:
It seems Travis is running `bundler` with the option `--deployment` enabled. Which makes brakeman command not available to the build. Need to fix that before move forward with this PR

PR #1154, opened 2017-04-13
Show a tantalising message if there's been no edits in the last week

henare:
Before this change the recent edits page showed a blank page if nothing
had happened in the last week. It now tries to entice the viewer into
contributing by reminding them that they'll be shown on the history page
if they make a contribution. It also links off to pages to help them do
that.

selection_001

### DIFF
--- a/app/views/home/history.html.haml
+++ b/app/views/home/history.html.haml
@@ -5,4 +5,8 @@
     .page-header.container
       %h1 All edits made in the last week
 
-= render partial: "layouts/history_list"
+- if @history.empty?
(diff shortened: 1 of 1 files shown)


equivalentideas:
Thanks @henare I made a few changes to try and make this simpler,
and also changed the research page to make it a clearer path through
to taking action.

See https://github.com/openaustralia/publicwhip/compare/96baadce35f6...8df5e62b7693

equivalentideas:
This is deployed 🚢

Issue #1144, opened 2017-04-03
Remove tests that compare pages with old PHP app


mario-areias:
We have many regression tests that send requests using the PHP URL format inherited from the original [publicwhip](https://bitbucket.org/publicwhip/publicwhip-v1/).  They are quite useful from a regression standpoint and capture many errors that otherwise would have passed.  However, they are a major barrier for people who want to do their first contribution to TVFY, particularly junior developers.

Here is a list of why they are a bit confusing:

- They use PHP URL format, not the Rails one.
```ruby
    it {compare_static("/division.php?date=2013-03-14&number=1&house=representatives")}
    it {compare_static("/division.php?date=2013-03-14&number=1&house=senate")}
    it {compare_static("/division.php?date=2013-03-14&number=1&house=representatives&display=policies", false, false, "_2")}
    it {compare_static("/division.php?date=2013-03-14&number=1&house=senate&display=policies", false, false, "_2")}
```

- There are too many of them. A superficial search shows us we have 118 of them. Mostly focused on the divisions page
```
 find spec/requests/ -name *.rb | xargs egrep compare_static | wc -l
```
- They test too much. I added a random string at the `app/views/divisions/index.html.haml` and then I ran the tests. This little change broke 61 tests. That's a bit overwhelming for such a small change.

```
Finished in 28.4 seconds (files took 3.3 seconds to load)
428 examples, 61 failures
```

- The error reporting is not good. In order to see the errors, one needs to use `git diff` to verify what has changed.
 ```
  61) DivisionsController#index 
      Failure/Error: raise "Don't match. Writing over file in spec/fixtures/static_pages. Do a git diff."
      
      RuntimeError:
        Don't match. Writing over file in spec/fixtures/static_pages. Do a git diff.
      # ./spec/html_compare_helper.rb:83:in `compare_text'
      # ./spec/html_compare_helper.rb:67:in `compare_static'
      # ./spec/requests/divisions_controller_spec.rb:71:in `block (3 levels) in <top (required)>'
```
- They are not easily reproducible. Due to their nature of saving the changes on the HTML files, if you run the tests twice, the tests will pass.

mario-areias:
Some suggestions in how to address that. For every new PR from now one, it is required to have tests. If they overlap with the request tests, we then delete the requests tests. It's a slow approach but it's the safest one, I reckon.

[...]

PR #1159, opened 2017-05-17
[WIP] Replace fixtures with factories and only load required data

equivalentideas:
This PR improves our test suit by:

* moving all the fixture data into factories, which makes our database cleaning much simpler and we don't need to add `.delete_all` hacks at the beginning of each spec file. This was a real gotcha that hit people trying to contribute tests (including me!).
* only calls in the required data for each of the static regression tests. This makes it much easier to understand what the test is doing, which is a step towards replacing these tests with more focused and understandable types of tests (see https://github.com/openaustralia/publicwhip/issues/1144 )
* makes loading the homepage not dependent of having data loaded, which makes some tests simpler and doesn't impact users. There's also a refactor of the view in there.
* adds TODOs suggesting improvements to the tests

It also slows down the test suit run by 15 seconds—but I think that's a reasonable price to pay for the increased clarity for developers, and as a stepping stone towards removing the static regression tests https://github.com/openaustralia/publicwhip/issues/1144

I did this while I was on the ✈️ , so it might be a bit crazy. I'm interested in your thoughts @mario-areias . Is this a win and a step towards https://github.com/openaustralia/publicwhip/issues/1144 in your opinion?

The other thing is, do you think this reduces clarity of the test setup for the regression tests because there's potentially base factory data coming through, and therefore could change behaviour without realising it?


equivalentideas:
There's currently some invalid factories that need to be fixed:

```
/Users/Luke/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/factory_girl-4.7.0/lib/factory_girl/linter.rb:12:in `lint!': The following factories are invalid: (FactoryGirl::InvalidFactoryError)

* electorate - Mysql2::Error: Duplicate entry '1' for key 'PRIMARY': INSERT INTO `electorates` (`id`, `name`, `main_name`, `house`, `created_at`, `updated_at`) VALUES (1, 'Warringah', 1, 'representatives','2017-05-17 07:46:27', '2017-05-17 07:46:27') (ActiveRecord::RecordNotUnique)
* policy_person_distance - Mysql2::Error: Duplicate entry '1-10552' for key 'dream_id_2': INSERT INTO `policy_person_distances` (`nvotessame`, `nvotessamestrong`, `nvotesdiffer`, `nvotesdifferstrong`, `nvotesabsent`, `nvotesabsentstrong`, `distance_a`, `distance_b`, `policy_id`, `person_id`, `created_at`, `updated_at`) VALUES (0, 0, 1, 0, 0, 0, 1.0, 1.0, 1, 10552, '2017-05-17 07:46:28', '2017-05-17 07:46:28') (ActiveRecord::RecordNotUnique)
* person - Mysql2::Error: Duplicate entry '10458' for key 'PRIMARY': INSERT INTO `people` (`small_image_url`, `large_image_url`, `id`, `created_at`, `updated_at`) VALUES ('http://www.openaustralia.org/images/mps/10458.jpg', 'http://www.openaustralia.org/images/mpsL/10458.jpg', 10458, '2017-05-17 07:46:28', '2017-05-17 07:46:28') (ActiveRecord::RecordNotUnique)
```

equivalentideas:
@mario-areias Thanks for these thoughtful comments, sorry it's taken me so long to get back to you.

> I think generally this a step in the right direction. The test setup is a bit more verbose now, but in the other hand I think it makes clear the dependency of each test. It's also easier to track down which tests will be impacted by any change on the factories.

Nice that's what I was hoping 👍 this is intended as a step towards removing the tests we can and making test setups isolated and consistent in their cleanup.

> It would be awesome if as part of this PR you could check if there are different tests testing the same thing (I had the impression this is happening, but I am not sure). Potentially that will help to decrease the number of tests and number of factories created.

While my head was in this, I was beginning to understand which tests were redundant, but it's still hard to understand it coming in cold. The problem is that lots of the tests still share a big setup, it's not clear which parts of the data created they're using. So you can't easily tell which tests use the same data to render the same pages.

By moving to the factories here, and only calling the required setup data per file, it's at least a little easier to understand what the test setup is, but still miles from being clear.

But I think the next step is to reorganise that setup into themes rather than make it based on data type. So you setup `create_tony_abbot_for_regression_tests` or `create_tony_abbot_with_policy_positions_for_regression_tests` and stuff like that, rather than `create_people_for_regression_tests`. I think that would make it easier to understand what's being used.

But...I don't feel like I can just back into this and continue to spike to make that next step.

@mario-areias do you think this would still be a positive change if I just write that up as a TODO in the file and explain it on https://github.com/openaustralia/publicwhip/issues/1144?

mario-areias:
@equivalentideas sorry for the late reply.

Yes, it's a positive change as long as we don't stop in there. I might have a deeper look at that in July, so we keep the ball rolling.

Awesome work!

equivalentideas:
> Yes, it's a positive change as long as we don't stop in there. I might have a deeper look at that in July, so we keep the ball rolling.

Nice @mario-areias thanks for that. I'll get the tests passing then and we can merge this.

PR #1167, opened 2017-11-18
Added a summary to the top of the policy page.

### LINKED ISSUE #1093: Add a summary of groupings, with anchor links and a bar graph visualisation, so I can quickly asses relative support for a policy
### Rationale

If a visitor wishes to assess relative support for a policy, they currently need to scan the entire page (which can be very long), or sometime even count the number of profile photos in each support grouping.

This change attempts to better support this assessment by offering a summary of the level of support for or against a position. The bar graph layout is here to support users with a preference for visual learning styles.

Each summary line is an anchor link to the grouping, to allow users to quickly jump to a controversial or interesting group.
### Layout

option 3c category filter and summary
### Notes
- The layout above includes features from #1090 (Add names...), #1091 (live text filter) and #1092 (party & house filters), but these aren't dependencies (a clean version is below)
- At small screen widths, each bar should ideally wrap underneath the relevant anchor link as shown below
  option 3c summary - desktop vs mobile layouts
#### MVP

This "clean" version below shows the concept without features covered in other issues.

option 2c summary no cat filter

Note, however, the change to the divisions anchor link ("4 divisions"), which has been reduced in size and changed in text. IMO it should have the same prominence as the summary links (reviewing particular divisions is quite an advanced step in reviewing a policy)

cofiem:
Followed @brtrx's suggestion, option 2e
See https://github.com/openaustralia/publicwhip/issues/1093#issuecomment-213099376
Closes #1093

### DIFF
--- a/app/assets/stylesheets/policies/_policies.scss
+++ b/app/assets/stylesheets/policies/_policies.scss
@@ -1,3 +1,4 @@
 @import "policies-list";
 @import "policies-new";
 @import "policy-voter-comparision";
+@import "policy-voter-comparision-summary";
--- a/app/assets/stylesheets/policies/_policy-voter-comparision-summary.scss
+++ b/app/assets/stylesheets/policies/_policy-voter-comparision-summary.scss
@@ -1,0 +1,58 @@
+.policy-comparision-summary {
+
+  .policy-comparision-summary-position {
+
+  }
+
+  .policy-comparision-summary-bars {
+    padding-top: 7px;
+
+    .policy-comparision-summary-bar {
+      height: 20px;
+      float: left;
+      margin-right: 1px;
+
+      &.very-strongly-for {
+        background-color: $color-very-strongly-for;
+      }
+
+      &.strongly-for {
+        background-color: $color-strongly-for;
+      }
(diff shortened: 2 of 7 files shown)


cofiem:
Example screenshot:

s

equivalentideas:
Great work implementing this complex displace @cofiem , however I'm not sure this is a feature that is really justified. See my comment on the issue https://github.com/openaustralia/publicwhip/issues/1093#issuecomment-345879327

cofiem:
If finding a particular MP, or group of MPs, is the most common goal, then yes I agree, this summary doesn't add much. 

However, there currently isn't any indication of the overall support/opposition to a policy. For me, in addition to finding out how my MP vote, I also want to see where the policy is at e.g. is there majority support? 

I suppose a quick scroll gives a similar overview to the chart. Perhaps going with the party/indep. support/opposition idea rather than just vote counts provides more useful information? The very strong/strong/moderate bars could be changed out for party groupings, with the same For/Against/Mixed/Never sets? That would give an excellent overview of who supports/opposes/doesn't care about the policy.

PR #1168, opened 2017-11-18
Modified policy list page to add subscriber count and last division date

### LINKED ISSUE #1111: Show how many people are subscribed to each policy in the list of policies
There are a few hundred subscriptions to policies on They Vote For You, but which policies are people subscribing to? Are we working on the ones that people are most interested in?

Knowing this will be useful to prioritise work on the policies and marketing efforts, and I think it will begin to bring us a little closer to passionate users of PlanningAlerts.

Showing how many people subscribe to each policy in this [list of policies](https://theyvoteforyou.org.au/policies) will make this clear to us in a transparent way that users can also see. I think that putting this forward will help to remind everyone that other people are using the site, giving them social confirmation for their interest. I really like the idea of pushing forward the humans using the site, which is something that we've been playing with in PlanningAlerts over the last year.

As a first iteration of this I'll try putting them down the left vertical line, which is pretty high up in the hierarchy of the block. I think this will be interesting information for people. Something like:

screen shot 2016-12-08 at 3 00 03 pm

cofiem:
Also added css media query for better layout on small screen.
Closes #1111

### DIFF
--- a/app/assets/stylesheets/policies/_policies-list.scss
+++ b/app/assets/stylesheets/policies/_policies-list.scss
@@ -2,6 +2,33 @@
   .policy-title {
     font-weight: normal;
     padding-left: 0;
+
+    .policy-details {
+      margin-top:7px;
+
+      .policy-most-recent-division {
+        &:before {
+          content: "\00a0\25CF\00a0";
+          color: #aaa;
+        }
+      }
+    }
+  }
+
+  .policy-detail-large-screen {
+    @media (max-width: $screen-md-min) {
+      display: none;
+    }
+  }
+
+  .policy-detail-small-screen{
+    @media (min-width: $screen-md-min) {
+      display: none;
+    }
+    &:before {
+      content: "\00a0\25CF\00a0";
+      color: #aaa;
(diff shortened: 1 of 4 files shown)


equivalentideas:
✨ Nice @cofiem could you add some screenshots of how this appears at different viewport sizes?

mlandauer:
@cofiem I'm going to merge this but revert the bit that adds the "last division..." bit because I'm not convinced that this is the right place for that information. My thinking is that policies are like a measure of someone's entire voting history on a particular subject so they're kind of static in time, sort of. The divisions are the micro-detail that happen yesterday or tomorrow. They have a very current timelines but with that comes a need for a lot of arcane knowledge to understand what they mean.

It's not in the [project design principles](https://github.com/openaustralia/publicwhip/blob/master/design_docs/principles.md) but the site should support people on a learning journey where they uncover how things work and are able to delve into more detail as they understand more but we are mindful of not overloading the person at the start with too much information that's not immediately relevant. This is why we need to be so careful about not adding things that are not strictly needed.

PR #1174, opened 2018-03-01
Don't require mysql for a development install

jamezpolley:
New developers on a mac may not have mysql running locally, and this
shouldn't strictly be needed for local dev. So rather than having this
requirement present, move it into the :production group that that it's
only required if Rails.env is set to production.

Tweak docs to split out the install of mysql from the rest of the dev
setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,22 +45,50 @@ Once the people data has been loaded you can start loading votes. These are scra
 
 ## Development
 
-If your machine is already set up to develop Rails applications with MySQL just
-carry out the following steps and you should be good to go. Developing with
-[Vagrant](https://www.vagrantup.com/) is also possible (see below) but was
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -45,23 +45,51 @@ Once the people data has been loaded you can start loading votes. These are scra
...
+### Development install
+
+```
+bundle install --without production
+```
+
+will install all the requirements except mysql and dalli. If you do want to use mysql for development, you can
+
+```
+gem install mysql
```
mlandauer: This won't install the specific version of mysql in the Gemfile.lock. So, I think it would be better if this said instead:
`bundle install` which should install everything

mlandauer:
There are painful small differences in databases so it's generally good practice to use the same database for local development as used in production. So closing

Issue #1164, opened 2017-08-22
policies.json in API should report last updated timestamp


jnothman:
Large parts of the tvfy data are static. Policies, however, can be changed. If a last-updated timestamp was provided for each policy in the API (presumably the max over `policies.updated_at` and any corresponding `policy_divisions.updated_at`), and particularly in the overview, this would make it much easier for users to capture changes incrementally, rather than download all the data.

henare:
We already display the last edit time on the policies index page: https://theyvoteforyou.org.au/policies

Adding this to the API seems like a good idea and it should be very straightforward :+1:

kevinjcliao:
Hi there! Found this project and am giving this a shot. How do I disable the requirement for an API key when I'm developing locally?

kevinjcliao:
I just disabled the require_key in the applicationcontroller for the API for now. I'll revert it when I commit.

mlandauer:
Seems like a sensible and straightforward addition

PR #1177, opened 2018-06-26
Add last_edited_at to policy JSON builder for APIv1

kevinjcliao:
Referencing #1164. I'm not sure if this is the right way to go about it, but I just added last_edited_at, which is what the policy controller calls for the website. Let me know if this is wrong or naive!


kevinjcliao:
Made a mistake with branches, sorry.

Issue #1180, opened 2018-08-31
Better explanation needed for voting tiers


MicaelaMicaela:
I've now had two users write in to ask for more information about what exactly the different voting tiers mean (i.e. what 'voted very strongly for' means as opposed to 'voted strongly for' etc). Currently, we don't have any direct plain English explanation for this on the site.

What we do have is a link that says "[Something not right?](https://theyvoteforyou.org.au/help/faq#stance)" next to each voting tier on each policy page and representative voting record (e.g. see [carbon price policy](https://theyvoteforyou.org.au/policies/3) and [Albo's voting record](https://theyvoteforyou.org.au/people/representatives/grayndler/anthony_albanese)). This link goes to our FAQ page where we give a **very brief** explanation of how our site works, which frankly isn't that helpful (and I fully acknowledge that I helped write it several years ago....).

I propose we change this link to something like "What does this mean?" and link to the following explanation, which is what I emailed in response to the queries:

> They Vote For You uses rankings like 'strongly' and 'very strongly' to take account of the fact that while our MPs and Senators can only vote Yes or No in any given [division](https://theyvoteforyou.org.au/divisions) (or formal vote), there are several divisions connected to each [policy](https://theyvoteforyou.org.au/policies) and not all those divisions have the same significance. So, when you connect a division to a policy, you have to select how someone who supported that policy would vote. The options are:
>
> * Yes (strong)
> * Yes
> * No (strong)
> * No
>
> We give the strength option as a way to distinguish between important and less important votes. For example, for our [marriage equality policy](https://theyvoteforyou.org.au/policies/1), a vote to pass a bill that would legalise same-sex marriage is clearly an important vote and so worth the "strong" marker. But a vote on recognising that it's the National Year of Action on Marriage Equality is less important since it doesn't actually change anything and is simply an awareness raising exercise. Both these votes relate to the policy and so have been connected to it, but with different levels of significance.
>
> For an MP or Senator to be marked as "voted very strongly for" a particular policy, they need to have almost always voted in a way that supports the policy (they may have voted against it once or twice, but usually only in a less important vote) and they need to have been present for almost every vote (too many absences will count against them and may even push them into the "voted against" categories since in our Parliament, not voting ultimately has the same effect as voting against something).
>
[...]

PR #1181, opened 2018-09-06
Add better explanation for voting tiers #1180

kevinjcliao:
#1180


kevinjcliao:
oh shoot. this includes a change from my other PR. I'll resubmit. Hang on.

PR #1183, opened 2018-09-06
Add better explanation for voting tiers #1180

### LINKED ISSUE #1180: Better explanation needed for voting tiers

kevinjcliao:
Fixes #1180

### DIFF
--- a/app/views/help/faq.html.haml
+++ b/app/views/help/faq.html.haml
@@ -23,7 +23,7 @@
 
     %ul
       %li= link_to "What are Policies and how do they work?", "#policies"
-      %li= link_to "How did you decide whether someone supports a policy or not?", "#stance"
+      %li= link_to "How did you decide the degree to which someone supports a policy?", "#stance"
       %li= link_to "What are “Rebel Voters”?", "#rebel"
       %li= link_to "What is a “Free Vote”?", "#free"
       %li= link_to 'What do the "attendance" figures mean?', "#attendance"
@@ -51,7 +51,7 @@
     %h2#policies What are Policies and how do they work?
     = render "help/faq/policies"
 
-    %h2#stance How did you decide whether someone supports a policy or not?
+    %h2#stance How did you decide the degree to which someone supports a policy?
     = render "help/faq/stance"
 
     %h2#rebel What are “Rebel Voters”?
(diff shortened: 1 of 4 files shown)


jamezpolley:
Closing this; the changes here have been expanded on in #1189.

Thanks @kevinjcliao!

PR #1227, opened 2020-10-20
Update ruby version to the minimum required for bundler to be install…

jamezpolley:
…ed by our ansible roles


jamezpolley:
On second thoughts: 2.5.2 still has other issues that make it hard to use (see https://bundler.io/blog/2019/05/14/solutions-for-cant-find-gem-bundler-with-executable-bundle.html)

I'm going to jump to 2.6.3 instead as that has a version of Rubygems that fixes an issue with Bundler.

PR #1295, opened 2021-12-14
Fixed alignment of top paragraph in members policies view.

Taz17:
This PR fixes issue #1095 by adding a .remove-padding class to the `<p>` tag.

### DIFF
--- a/app/assets/stylesheets/policies/_policy-voter-comparision.scss
+++ b/app/assets/stylesheets/policies/_policy-voter-comparision.scss
@@ -130,3 +130,7 @@
 *:hover > .anchor {
   display: block;
 }
+
+.remove-padding {
+  padding: 0px 0px;
+}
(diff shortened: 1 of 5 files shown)


mlandauer:
Thanks @Taz17 for looking at this. Your solution looks good to me. A couple of things: you'll need to make sure that all the tests pass as part of a pull request. To do that the expected results of certain pages will change so those will need updating. Also, for css / design changes in PRs it's really helpful to include before and after screenshots. Could you do that here as well? Thank you!

Taz17:
Before and after pictures:

Before:

tvfy-before

After:

tvfy-after

mlandauer:
Thanks for doing the screenshots @Taz17 . That helps hugely. Now from the looks of it, the left hand side of the photo doesn't line up with the left hand side of that first paragraph. Take another look at the original bug report.

Taz17:
Ah, apologies I had played around with the zoom to get a good screenshot, so that must have been it. Here are better before and after images:

before:

tvfy-before (2)

after:

tvfy-after (2)

mlandauer:
Great. Thanks! 👍🏼

Issue #1289, opened 2021-11-30
help/faq: a number of external links are broken


bobsbarrels:
There are links to external webpages that no longer exist:

This is the list found so far, just eyeballing "faq", though there may be more.

---

https://theyvoteforyou.org.au/help/faq#division-occur

Most decisions in Parliament are made **‘on the voices’** and not by division. 

broken: https://peo.gov.au/understand-our-parliament/voting-in-the-chambers.html

---

https://theyvoteforyou.org.au/help/faq#rebel

voting against the **party whip**. 

broken: https://peo.gov.au/understand-our-parliament/party-whip.html

---

https://theyvoteforyou.org.au/help/faq#attendance

Link from **pair** is broken:

broken:  https://www.aph.gov.au/Help/404?item=%2fabout_parliament%2fsenate%2fpowers_practice_n_procedures%2fodgers%2fchap1107&user=extranet%5cAnonymous&site=website

![Screenshot from 2021-12-01 07-44-31](https://user-images.githubusercontent.com/26145487/144125205-71d3f6cd-76ed-4152-a286-5d32c05b0c5b.png)

---

https://theyvoteforyou.org.au/help/faq#tied-division

tied it is **considered to have failed** because

broken: https://peo.gov.au/understand-our-parliament/voting-in-the-chambers.html

bobsbarrels:
If correct link is provided I'm happy to fix this :)

bobsbarrels:
Actually this is the tip of an iceberg. Collecting bunch of broken external links ... TBA.

katska:
The information is now at https://www.aph.gov.au/About_Parliament/Senate/Powers_practice_n_procedures/Odgers_Australian_Senate_Practice/Chapter_11 although not in a usefully separate section as before http://web.archive.org/web/20170403002105/http://www.aph.gov.au/About_Parliament/Senate/Powers_practice_n_procedures/odgers/chap1107

bobsbarrels:
> The information is now at https://www.aph.gov.au/About_Parliament/Senate/Powers_practice_n_procedures/Odgers_Australian_Senate_Practice/Chapter_11 although not in a usefully separate section as before http://web.archive.org/web/20170403002105/http://www.aph.gov.au/About_Parliament/Senate/Powers_practice_n_procedures/odgers/chap1107

[...]

PR #1296, opened 2021-12-14
Fix faq link pairs #1289

bobsbarrels:


bobsbarrels:
As discussed: https://github.com/openaustralia/publicwhip/issues/1289

Would mind review from @katska as I kind of guessed that the PEO links mapped, all in this commit:
https://github.com/openaustralia/publicwhip/pull/1296/commits/1a412d383071b7c52733c88facb3551ed4d7d745

(The other one was per her advice :) )

mlandauer:
Thanks @bobsbarrels. I had to update the target result for the regression tests to match your changes. For future reference you can do the same by running `bundle exec rake` and eye-balling the result

PR #1300, opened 2021-12-15
Fixed  calculation of attendance in divisions

### LINKED ISSUE #1202: 150% Attendance at division
In this division, it says PHON had 150% attendance:

https://theyvoteforyou.org.au/divisions/senate/2018-06-27/5

Taz17:
This PR fixes #1202.

The maximum attendance can now only be 100.0

### DIFF
--- a/app/models/division.rb
+++ b/app/models/division.rb
@@ -118,7 +118,7 @@ class Division < ApplicationRecord
 
   # Returns nil if otherwise we would get divide by zero
   def attendance_fraction
-    total_votes.to_f / possible_votes if possible_votes.positive?
+    [total_votes.to_f / possible_votes, 100.0].min if possible_votes.positive?
   end
 
   def edited?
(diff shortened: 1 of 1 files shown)


mlandauer:
This PR is just a workaround / hack. It's not fixing or addressing the underlying issue that's causing the problem in #1202. In theory attendance as calculated should never be above 100%. It clearly is, something fundamental is going wrong. So, I'm going to close this without merging.

I suggest first trying to reproduce the problem in #1202 locally. This probably will involve loading some production data for that day which will be an interesting exercise in its own right ;-)

If you can reproduce the problem then you're probably really close to solving it. Good luck!

Taz17:
@mlandauer Ok i'll try to reproduce it locally, how can I get my hands on that production data?

mlandauer:
> @mlandauer Ok i'll try to reproduce it locally, how can I get my hands on that production data?

https://github.com/openaustralia/publicwhip/#australia-1 should hopefully point you in the right direction. You'll also need to run rake tasks to update the caches after new divisions are loaded

PR #1301, opened 2021-12-16
Rename fixture files to allow git clone on Windows

cofiem:
The character '?' is not allowed in paths on Windows. This means this repo as it is cannot be cloned on Windows.

Git gives this error: `error: invalid path 'spec/fixtures/static_pages/search?query=0000.html'`.

This PR renames the files to replace '?' with '__'. It also makes the same replacement when loading the files for tests.

There are also some string replacements of "\r\n" with "\n" to ignore newline differences when running the tests on Windows.


mlandauer:
Looks good to me. Thanks @cofiem! 🎉

Issue #1304, opened 2021-12-20
Periodically check that all links in markdown for divisions are valid


mlandauer:
For instance on this division https://theyvoteforyou.org.au/divisions/representatives/2007-08-07/1 there is a link for "Jenny Macklin" that uses an old (php style) link to theyvoteforyou which we are no longer redirecting because it's long-since been phased out in other areas and was only used briefly before the public launch of the site.

We could just rewrite those links but it makes sense to instead generalise the solution a little so that we check all links in those descriptions to make sure they are up to date.

I think an initial "proof of concept" could be a rake task that just outputs the urls of problematic divisions to the console (with maybe the specific links in question that are broken).

If that proves to be useful maybe then we could instead something that is more integrated into the app for a normal user. For instance we could run a cron job once per week that checks all the links and then flags broken ones by marking them in the database. Then, we could show the user that there is a broken link on the division detail page. Also, we could have a separate page which lists all the divisions with broken links.

mlandauer:
@Taz17 could you please pick this one up? Please only do the simple version, the "proof of concept" to make a rake task to output the problematic division urls and the specific links that are the problem. I want the solution to not make more external http requests than required and to not depend too specifically on the markup format (e.g. use the markdown parser we're already using to do the heavy lifting)

mlandauer:
The idea is then we can give that list to @mackaymackay to go through and fix

Taz17:
Closing this issue since its done.

PR #1309, opened 2021-12-23
Link validation rake task

Taz17:
Basic proof of concept to check for old php links. PR is targeted towards issue #1304  

Currently implemented to search for urls using regex rather than using `redcarpet` the markdown parser.

Tested locally and is able to find urls containing `php` in them.

Current output of this rake task:

image

Note: that https://theyvoteforyou.org.au/divisions/senate/2014-02-13/1 does not have this issue since I ran it locally without production data.


mlandauer:
@Taz17 thanks for looking at this.

Perhaps I didn't make it clear enough in the original issue #1304 but I want the links to be validated by doing actual web requests to them to see if they work or error. So, I would expect the validator to extract the url, do a web request to that url. If it returns a 200 code (successful) then it's good. If it redirects the redirect should be followed and that url should be tested.

Also, where the same link comes up multiple times (in different divisions, for example) I want only one web request to be made. This is because there are some links that are used many, many times and I don't want a burst of traffic coming from us that isn't necessary.

The use of markdown renderer is necessary because these formats can be very flaky and use of regular expressions should be avoided unless really necessary. See for example https://blog.codinghorror.com/regular-expressions-now-you-have-two-problems/

What I suggest doing instead is using the markdown renderer to convert the text to html and then parsing the resulting html using [nokogiri](https://nokogiri.org/). Then you can search for links using nokogiri and extract the urls. That way of doing things should be very reliable and easy to understand though it will be somewhat computationally wasteful but I wouldn't worry about that much here. Clarity, reliability and maintainability trump other things.

lib/tasks/application.rake
```diff
@@ -121,4 +121,31 @@ namespace :application do
     Rails.logger = ActiveSupport::Logger.new($stdout)
     Rails.logger.level = 1
   end
+
+  namespace :links_valid do
+    desc "Checks that all links in markdown for divisions are valid"
```
mlandauer, line 126: The description doesn't actually accurately reflect what this is doing

lib/tasks/application.rake
```diff
@@ -121,4 +121,31 @@ namespace :application do
...
+    task divisions: :environment do
+      base_url = "https://theyvoteforyou.org.au"
+
+      Division.all.each do |division|
```
mlandauer, line 130: Use `find_each` instead of `find` here https://apidock.com/rails/ActiveRecord/Batches/ClassMethods/find_each

mlandauer:
@Taz17 in my eagerness to give you feedback and keep you moving I didn't say a couple of important things.

Thank you so much for diving in there and getting something quickly up and running.

Secondly I also really like that you made a "draft" PR to share your work in progress even though you knew it wasn't probably complete. That was so helpful because I could provide you some big picture feedback quickly and it gave me a really good idea of where you are up to. So thank you for that!

lib/tasks/application.rake
Taz17: Updated description in the last commit.

Taz17:
Run `bundle exec rake application:links_valid:divisions` to check for links validitiy in divisions summary.

Only outputs once a broken link is found.

Output looks like this:

image

To be done:
1. Split URL into domain and path since Net::HTTP.get only requests the homepage of every domain.
2. Check for re-direction and follow that URL.

Taz17:
As of the most recent commits, the rake task now checks for all broken links in markdown files found in a divisions description. Only one request is made if a URL appears more than once.

mlandauer:
@Taz17 before I look through your changes can you make sure that all the CI tests pass. I can see that there's some failures which are caused by rubocop which is a ruby style / best practises checker. You can run it locally by doing `bundle exec rubocop`.

Taz17:
@mlandauer Ah yes, it was the rubocop that made the build fail. You should be able to take a look at it now.

lib/tasks/application.rake
```diff
@@ -121,4 +121,43 @@ namespace :application do
...
+          summary_in_html = md.render(wiki_motion.description)
+          parsed_data = Nokogiri::HTML.parse(summary_in_html)
+          url_hash_table = Hash.new(0)
+
+          tags = parsed_data.xpath("//a")
+          tags.each do |tag|
+            url_from_page = tag[:href]
+            url = url_from_page
+
+            url = "http://#{url}" if !url[%r{\Ahttp://}] && !url[%r{\Ahttps://}] # adds http to the start of the URL to avoid failure
```
mlandauer, line 144: Can you add a comment explaining a bit more why this is being added. Is it because of the way markdown can interpret urls that are not complete?

lib/tasks/application.rake
```diff
@@ -121,4 +121,43 @@ namespace :application do
     Rails.logger = ActiveSupport::Logger.new($stdout)
     Rails.logger.level = 1
   end
+
+  namespace :links_valid do
+    desc "Checks the validity of links in division summary"
+    task divisions: :environment do
+      base_url = "https://theyvoteforyou.org.au"
```
mlandauer, line 128: Ideally we'd get this from a configuration rather than hardcoding it but it looks like we don't actually have a configuration that we can trivially use. Other OAF projects have something like this but not here unfortunately yet.

lib/tasks/application.rake
```diff
@@ -121,4 +121,43 @@ namespace :application do
...
+
+            url = "http://#{url}" if !url[%r{\Ahttp://}] && !url[%r{\Ahttps://}] # adds http to the start of the URL to avoid failure
+
+            begin
+              if (url_hash_table[url]).zero?
+                url_hash_table[url] += 1
+                uri = URI(url)
+                Net::HTTP.get(uri)
+              end
+            rescue StandardError
```
mlandauer, line 152: If a particular URL that is broken appears on several division pages won't this logic just show the first one that it errors on rather than showing all the division pages that the broken url is on (while at the same time only doing one web request)?

lib/tasks/application.rake
```diff
@@ -121,4 +121,43 @@ namespace :application do
...
+              puts "---------------------------------------------------------------------------------------------"
+              puts "URL seems broken! URL: #{url_from_page}"
```
mlandauer, line 154: From a user point of view if I want to find out about broken links and I fix them division by division. I want to get all the broken links on a particular division in one block. Something like:
```
There are broken links in the description for division https://theyvoteforyou.org.au/divisions/senate/...:
   /foo/bar
   /anotherlink/to/somewhere else
```

lib/tasks/application.rake
```diff
@@ -121,4 +121,43 @@ namespace :application do
...
+              puts "URL found in the summary of this division: #{base_url}/divisions/#{division.house}/#{division.date}/#{division.number}"
```
mlandauer, line 155: Use url helpers for this so that you don't hardcode paths https://guides.rubyonrails.org/routing.html#path-and-url-helpers.

mlandauer (changes requested):

lib/tasks/application.rake
Taz17: Could you refer me to an OAF project with the appropriate configuration?

Taz17: Is this even necessary anymore? It is now removed after using URL helpers

mlandauer: I'm not sure off the top of my head. With the url helpers in normal views it gets the domain from the web request. For mailers it gets the domain from the configuration [here](https://github.com/openaustralia/publicwhip/blob/master/config/environments/production.rb#L69). I'm not totally sure what happens in a rake task off the top of my head

lib/tasks/application.rake
```diff
@@ -121,4 +123,45 @@ namespace :application do
...
+            url = "http://#{url}" if !url[%r{\Ahttp://}] && !url[%r{\Ahttps://}]
+
+            begin
+              if (url_hash_table[url]).zero?
+                url_hash_table[url] += 1
+                uri = URI(url)
+                Net::HTTP.get(uri)
+              end
+            rescue StandardError
+              broken_urls.append(url_from_page)
```
mlandauer, line 156: A more concise version is `broken_urls << url_from_page`

lib/tasks/application.rake
```diff
@@ -121,4 +123,45 @@ namespace :application do
...
+          parsed_data = Nokogiri::HTML.parse(summary_in_html)
+          url_hash_table = Hash.new(0)
+          broken_urls = []
+
+          tags = parsed_data.xpath("//a")
+          tags.each do |tag|
+            url_from_page = tag[:href]
+            url = url_from_page
+
+            # adds http to the start of the URL to avoid failure since web request will fail without http://
```
mlandauer, line 146: This comment still doesn't actually explain why urls are appearing without "http://"

lib/tasks/application.rake
```diff
@@ -121,4 +123,45 @@ namespace :application do
...
+        if wiki_motion
+          summary_in_html = md.render(wiki_motion.description)
+          parsed_data = Nokogiri::HTML.parse(summary_in_html)
+          url_hash_table = Hash.new(0)
+          broken_urls = []
+
+          tags = parsed_data.xpath("//a")
+          tags.each do |tag|
+            url_from_page = tag[:href]
+            url = url_from_page
```
mlandauer, line 144: why are you assigning this to `url_from_page` first and then `url`?

lib/tasks/application.rake
```diff
@@ -121,4 +123,45 @@ namespace :application do
...
+
+            # adds http to the start of the URL to avoid failure since web request will fail without http://
+            url = "http://#{url}" if !url[%r{\Ahttp://}] && !url[%r{\Ahttps://}]
+
+            begin
+              if (url_hash_table[url]).zero?
```
mlandauer, line 150: I can see how this will avoid doing multiple web requests when there a single url appears multiple times in a particular division. However, I don't think this will handle the (more likely) case where a particular url appears multiple times across many divisions. We want to make sure that even in that case we don't do multiple web requests to the same url.

mlandauer (changes requested):

lib/tasks/application.rake
Taz17: Just in the case that a URL is missing the http://. As an edge case.

lib/tasks/application.rake
Taz17: `url` is used when making the web request as it will always have the `http://` Infront of it even when `url_from_page` doesnt

`url_from_page` will not always have the `http://`. Additionally, its used when printing the broken URL out. You can simply copy and ctrl + f to search for the broken URL (since it does not change but `url` does)

lib/tasks/application.rake
mlandauer: Yes I understand but it still doesn't explain **why** that can happen. A url should always have "http://" or "https://" in front of it. The comment should explain why a non-valid url can appear in the markdown and if it isn't a valid url then why should it be hacked to be valid?

lib/tasks/application.rake
mlandauer: That makes sense thanks.

lib/tasks/application.rake
Taz17: Working on this now.

lib/tasks/application.rake
Taz17: I thought of it as an edge case. If you are saying that it should and will always have "http://" or "https://" then its better to be removed.

mlandauer: It's the markdown renderer that is rendering the html with the links so it seems very unlikely to me that it would generate urls without "http(s)" if front of it and for those urls to work correctly in a browser but not be valid urls. So, yes, it makes sense to me to remove this. Maybe we'll find some other edge cases when we actually run this in production.

mlandauer:
@Taz17 is this ready to review again?

Taz17:
@mlandauer Yes!

lib/tasks/application.rake
```diff
@@ -122,4 +124,46 @@ namespace :application do
...
     Rails.logger.level = 1
   end
+
+  namespace :links_valid do
+    desc "Checks the validity of links in division summary"
+    task divisions: :environment do
+      include PathHelper
+
+      md = Redcarpet::Markdown.new(Redcarpet::Render::HTML)
+      broken_url_constant = 999
```
mlandauer, line 134: Having to do something like this is a bit of "code smell" https://en.wikipedia.org/wiki/Code_smell. I guess more deeply you're using `url_hash_table` to do two separate things. It's recording whether something's cached and whether it's a valid url.

lib/tasks/application.rake
```diff
@@ -122,4 +124,46 @@ namespace :application do
...
+                Net::HTTP.get(uri)
+              elsif url_hash_table[url] == broken_url_constant
+                broken_urls << url
+              end
+            rescue StandardError
+              url_hash_table[url] = broken_url_constant
+              broken_urls << url
+            end
+          end
+          puts "There are broken links in the description for division #{division_url_simple(division)}"
```
mlandauer, line 161: Won't this get displayed even if there aren't any broken links on a page?

mlandauer (approved):
Thanks for going through this again. I hope it's not too frustrating going over this again and again. The code could still be improved to make it simpler and more readable (and less likely to have bugs). However, in the interest of keeping moving I'm going to merge this as is. Thanks for doing this!

lib/tasks/application.rake
Taz17: Fixed in PR #1318

lib/tasks/application.rake
Taz17: Fixed in PR #1318

PR #1318, opened 2022-01-18
Fixed code smell mentioned in previous PR conversation

Taz17:
1. Refactored `url_hash_table` to `visited_url`.
2. `visited_url` now has 1 use only and that is to cache the visited urls.
3. `broken_urls_table` now stores the broken urls
4. Only prints when there are broken urls.
5. Constant removed since its no longer in use.


mlandauer:
Hi @Taz17 I went through and made a few changes to the code after it was merged. So, I think that makes this PR not relevant anymore. It might be useful to cast your eye over those changes. Hopefully there's stuff in there you can learn from.

Thanks again!

Issue #1000, opened 2014-12-03
add on page search to policies page (with js)


equivalentideas:
This would enable users to quickly filter the policies list to find ones they are interested in. It can be quite difficult to scroll through the very long list of policies and pinpoint one of interest. A simple search form that filtered results (based on title initially) would make it more useable.

People are making searches from the policies page through the site search:

![screen shot 2014-12-03 at 11 08 07 am](https://cloud.githubusercontent.com/assets/1239550/5273535/bf0d51d6-7adc-11e4-85b0-a055f909f960.png)

mlandauer:
Yes to this a hundred fold! We have a lot of policies and this is the quickest way we could make that page more easy to use

PR #1332, opened 2022-02-10
Filtering policies based on name using JavaScript.

Taz17:
Relates to issue #1000 

Here is a gif of what it looks like:

js_search_gif

This is a draft PR for the following reasons:
1. Placeholder in the input box for search in my opinion could be better.
3. Design is not yet 100% responsive on small devices (phones). You can see in the left most screen in the image that the search box and the sorting of the policies are stacked on top of each other.
4. Styling is currently done in-line and not in a stylesheet
5. Javascript in the partial loads within the body (not at the end) of the HTML document. (I dont know if this is acceptable or not)

Screenshot from 2022-02-10 13-49-08

Would like feedback on this as well as suggestions on an appropriate layout of saerch+dropdown in the phone view.

@katska @mlandauer @tseyiokorodudu


Taz17:
Will update the regression tests and rubocop down the line as this is not ready for merging.

mlandauer:
Thanks @Taz17 I really like where this is going. Some early thoughts / feedback in no particular order:
* Love the feedback as you type. Can't wait to see this working in production
* In the narrow mobile view it makes sense to me that the two things should be stacked. However, I would expect them to be stacked the other way around as the filter is strong connected to the list of policies
* I agree the wording of placeholder in the input box isn't quite right. I'm not sure what is better, though. I'll have a think. Maybe "Search policies"?
* It needs some kind of sane fallback if javascript isn't enabled.
* I see in the commit history you've done repeated merges of master into your branch to get the latest changes. Before you publish your work as a PR you can use "git rebase" instead to rebase your work on to the current tip of master which gives a much cleaner and clearer version history.
* Try to do as much of the layout and styling using out of the box bootstrap
* Put your javascript in the assets directory rather than inline on the page. That way it will get minified and included in the standard js file so it doesn't need to get reloaded on every page load.
* I haven't looked closely but some of the stuff you're doing in your js you could be doing with jquery. I know it's very old school but it still has it's place. 
* It's good practise to try to keep your tests passing on every commit. It keeps the changes for a particular bit of code to go alongside the changes in the tests.

Thanks for putting this up as a PR before you considered it finished. Hope this feedback is helpful.

Taz17:
Here is the to-do list:
- [ ] Update placeholder text
- [ ] Place filter on top of the input in the mobile view
- [ ] Use out of the box bootstrap
- [ ] Place javascript in `/assets` directory
- [ ] Attempt a re implementation using jQuery

PR #1347, opened 2022-02-21
View - Social card1 visual design

tseyiokorodudu:
In this PR is the code responsible for social card 1 view. This includes the card haml and scss files along with an update to the members_controller file.

### DIFF
--- a/app/assets/stylesheets/card/_card.scss
+++ b/app/assets/stylesheets/card/_card.scss
@@ -1,0 +1,53 @@
+.card-container {
+    font-family: 'Fira Sans', sans-serif;
+    width: 600px;
+    height: 350px;
+    background-color: $color-lemon-light;
+}
+
+.logo {
+    font-size: 18px;
+    background-color: $color-lemon;
+    padding: 0.5em 1.75em;
+}
+
+.image-card {
+    display: grid;
+    grid-template-columns: 30% 1fr;
+    grid-column-gap: 25px;
+    grid-template-areas: 'a b';
+    margin: auto;
+    width: 90%;
+    padding-top: 1.4em;
+    padding-bottom: 1.4em;
+}
+
+.member-image {
+    grid-area: a;
+    img {
+        width: 150px;
+        height: 200px;
+    }
+}
+
+.fitin {
+    overflow: hidden;
+    max-height: 200px;
+}
+
+.heading-text {
+    grid-area: b;
+    color: #262626;
+    font-size: 36px;
+    line-height: 1.3;
+}
+
+.disclaimer {
+    padding-left: 3em;
+    margin-top: 14px;
+    font-size: 10px;
+}
+
+body::-webkit-scrollbar {
+    display: none;
+}
--- a/app/assets/stylesheets/card/_card.scss
+++ b/app/assets/stylesheets/card/_card.scss
@@ -1,17 +1,17 @@
-.card-container {
+.socialmedia-card {
     font-family: 'Fira Sans', sans-serif;
     width: 600px;
     height: 350px;
     background-color: $color-lemon-light;
 }
 
-.logo {
+.socialmedia-card .logo {
     font-size: 18px;
     background-color: $color-lemon;
     padding: 0.5em 1.75em;
 }
 
-.image-card {
+.socialmedia-card .image-card {
     display: grid;
     grid-template-columns: 30% 1fr;
     grid-column-gap: 25px;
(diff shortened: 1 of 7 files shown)


app/views/card/policy_with_member_card.html.haml
```diff
@@ -0,0 +1,33 @@
...
+                    .heading-text
+                        = policy_agreement_summary(@policy_person_distance, with_person: true, with_policy: true)
+    %p.disclaimer 
+        %span.subtext
+            = @link_to_policy
+
+
+
+:javascript
+    window.addEventListener('DOMContentLoaded', (event) => {
```
mlandauer, line 19: I think some comments in the code here to say why this javascript code is here and what is doing would be good

app/views/layouts/card_layout.html.haml
```diff
@@ -0,0 +1,29 @@
+
+!!! 5
+%html.no-js(lang="en")
+  %head
+    %meta(charset="utf-8")
+    %meta(http-equiv="X-UA-Compatible" content="IE=edge")
+    %meta(name="viewport" content="width=device-width, initial-scale=1")
+    %meta{:name => "robots", :content => "noindex"}
+    - if Settings.facebook_admins
```
mlandauer, line 9: Does this need to be here? As you're using a separate layout for the card then there's no reason to have this facebook stuff here.

app/views/layouts/card_layout.html.haml
```diff
@@ -0,0 +1,29 @@
...
+      - if content_for? :title
+        = yield :title
+        = "— #{Settings.project_name}"
+      - else
+        #{Settings.project_name} — How does your MP vote?
+
+    = stylesheet_link_tag "//code.cdn.mozilla.net/fonts/fira.css"
+    = stylesheet_link_tag "application"
+
+    <!--[if lt IE 9]>
```
mlandauer, line 22: The IE support is just needed as the only browser that's going to need to see this in the specific browser that's doing the screenshotting

app/assets/stylesheets/card/_card.scss
```diff
@@ -0,0 +1,53 @@
...
+    margin: auto;
+    width: 90%;
+    padding-top: 1.4em;
+    padding-bottom: 1.4em;
+}
+
+.member-image {
+    grid-area: a;
+    img {
+        width: 150px;
```
mlandauer, line 28: This is upscaling the image beyond its natural size so it's going to be blocky. For the purposes of this PR leave it as is but I do think we need to tackle that properly after this gets merged and figure out a pipeline for getting higher resolution images.

app/views/card/policy_with_member_card.html.haml
```diff
@@ -0,0 +1,33 @@
+.card-container
+    %p.logo 
+        %span.subtext They Vote For You 
+    .row
+        .col-sm-12 
+            .image-card 
+                .member-image 
+                    %img{:src => @member.large_image_url}
```
mlandauer, line 8: You can use the more modern ruby 1.9 syntax here `%img{src: @member.large_image_url}`

mlandauer (changes requested):
Thanks for getting this PR together. I appreciate the work you've put into it. There are a few issues that it would be good to try to fix up before this is merged. None of them are particularly big things I think. The image size issue we can leave till after this PR.

Let me know if you have any questions or any of my feedback didn't make sense! :-)

mlandauer:
I'm not sure why the CI (travis) didn't run over this PR. Hmmm..

app/controllers/members_controller.rb
```diff
@@ -62,7 +62,13 @@ def policy
     # If policy_person_distance doesn't exist then return a 404
     @policy_person_distance = @member.person.policy_person_distances.find_by!(policy: @policy)
 
-    render "policies/show_with_member"
+    return render "policies/show_with_member" if params[:card].nil?
+
+    @link_to_policy = "How is this worked out? #{request.original_url}"
```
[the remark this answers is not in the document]
tseyiokorodudu: Currently working on this.

mlandauer:
LGTM! Thanks @tseyiokorodudu 🎉

mlandauer:
This is now rolled out to production. I had to fix up a couple of things - it turned out the javascript compressor used in production wasn't setup to support ES6 syntax. That was config change in a2218169ca4a354518aa9c965219ff37ebd6693a. Also the change to move the javascript to its own assets file had to be reverted because it was actually getting included in the main `application.js` and so was running on every page causing errors.

PR #1353, opened 2022-02-23
Social card1 fixes

Taz17:
Fixes the changes made when javascript was moved from its own assets file to the .haml file itself


mlandauer:
Thanks!

mlandauer:
This change is now deployed to production

Issue #1340, opened 2022-02-16
Install chrome webdriver to server.


Taz17:
The chrome webdriver is needed so that we can capture screenshots of the social sharing cards mentioned in issue #1316

mlandauer:
Take a look at https://github.com/openaustralia/infrastructure/. I'd love to see a PR against that repo for any changes required to the infrastructure. I think you should be able to test most of it out locally. The only thing that would be hard to test would be an increase in disk space, so that would be something I probably would need to do.

Taz17:
Since you mentioned a gem that installs chrome driver once the gem is installed. I did some digging, and found that in the [multiple_cianywhere](https://github.com/planningalerts-scrapers/multiple_cianywhere) it uses `capybara` and `selenium-webdriver`. I tried to run it on a laptop I have without the webdriver installed by me and it did not run (driver not found exception)

Turns out morph chromedriver is installed as part of the build process for every docker container. Mentioned in the [documentation](https://morph.io/documentation/scraping_javascript_sites)

mlandauer:
This was fixed in https://github.com/openaustralia/infrastructure/commit/1cd955f8397b77d9dbc0f1dc8b88469fd6b6cddb

Issue #1338, opened 2022-02-15
Programmatically screenshotting and saving social media sharing cards.


Taz17:
To capture and screenshot the social media cards programmatically we will be using a headless browser to do so.

Ideally this would be in a rake task so that it can be run frequently to make sure every new social card is captured.

Taz17:
With this implementation comes the issue of where to save the screenshots of the cards. Saving the cards to the VM's harddisk is an option but I don't believe it would be sustainable as this could take up too much memory and slow down the machine. @mlandauer any ideas as to how to store the screenshots?

mlandauer:
> With this implementation comes the issue of where to save the screenshots of the cards. Saving the cards to the VM's harddisk is an option but I don't believe it would be sustainable as this could take up too much memory and slow down the machine. @mlandauer any ideas as to how to store the screenshots?

There's two separate issues here:
1. How much disk space are all the rendered images going to take up?
2. How much memory is needed to generate the images and how long will it take?

The first part isn't much of an issue I think. Disk space is pretty cheap. We can add an extra network mounted disk on the VM without too much trouble. We do something similar on Right To Know which has a lot of space on the machine for storing all the FOI documents. Nevertheless it would be extremely helpful if you could put together an estimate of how much disk space would be needed.

For the second part this could potentially be done in the middle of the night if it doesn't take too long and doesn't require our VM to be spec'd significantly differently. Again please figure out an estimate of the memory needed. I asked for this before. It's really important.

Taz17:
I can give you for an maximum amount of storage required for the screenshots of members and their vote for a particular policy. Their social card can be seen in issue #1339 (600x350 pixels)

Storage for all members and their vote on a policy = no. of all people * no. of all policies * no. bytes for 1 image.
= 227 * 226 * 65931 bytes
= 3382392162 bytes
= 3.38 gigabytes.

As for the other cards i cannot provide and estimate as I don't how many they are

Taz17:
Well given that storing to disk is a valid option, is the naming pattern `MemberID_PolicyID` sufficient?

eg. `production-screenshots/10001_1` for Tony Abbott on poilcy 1.

I assume it will be as it'll be unique and easy to query/search for.

mlandauer:
> As for the other cards i cannot provide and estimate as I don't how many they are

Why can't you provide an estimate of the other cards? What information are you missing? Don't you have an estimate of the number of people and the number of policies. Isn't that enough?

[...]

PR #1356, opened 2022-02-23
Screenshot rake task (clean git history)

Taz17:
This PR adds the code required to visit the pages of social media cards and capture screenshots of the webpage. https://github.com/openaustralia/publicwhip/issues/1338

Once https://github.com/openaustralia/publicwhip/issues/1340 is done we can add `task("application:generate:cards").invoke` to the daily rake task

### DIFF
--- a/app/lib/card_screenshotter/members.rb
+++ b/app/lib/card_screenshotter/members.rb
@@ -14,7 +14,7 @@ module CardScreenshotter
       PolicyPersonDistance.find_each do |ppd|
         person = ppd.person
         policy = ppd.policy
-        url = "http://#{ActionMailer::Base.default_url_options[:host]}#{person_policy_path_simple(person, policy)}?card=true"
+        url = "https://#{ActionMailer::Base.default_url_options[:host]}#{person_policy_path_simple(person, policy)}?card=true"
         file_name = person_policy_path_simple(person, policy).gsub("/", "_")
         file_name = "#{file_name}.png"
 
--- a/app/lib/card_screenshotter/members.rb
+++ b/app/lib/card_screenshotter/members.rb
@@ -5,7 +5,6 @@ include Rails.application.routes.url_helpers
 module CardScreenshotter
   class Members
     def self.update_screenshots
-      include PathHelper
       driver = CardScreenshotter::Utils.open_headless_driver
 
       card_width = 600
--- a/app/lib/card_screenshotter/members.rb
+++ b/app/lib/card_screenshotter/members.rb
@@ -22,5 +22,13 @@ module CardScreenshotter
       end
       CardScreenshotter::Utils.close_driver(driver)
     end
+
+    def self.get_save_path(person)
+      member = person.latest_member
+      house = member.house.downcase
+      constituency = member.constituency.downcase
+      name = member.first_name.concat("_", member.last_name).downcase
+      "public/cards/people/#{house}/#{constituency}/#{name}/policies"
+    end
   end
 end
--- a/app/lib/card_screenshotter/members.rb
+++ b/app/lib/card_screenshotter/members.rb
@@ -15,7 +15,8 @@ module CardScreenshotter
         person = ppd.person
         policy = ppd.policy
         url = "http://#{ActionMailer::Base.default_url_options[:host]}#{person_policy_path_simple(person, policy)}?card=true"
-        file_name = "#{person.id}_#{policy.id}.png"
+        file_name = person_policy_path_simple(person, policy).gsub("/", "_")
+        file_name = "#{file_name}.png"
 
         image = CardScreenshotter::Utils.screenshot(driver, url, card_width, card_height)
         CardScreenshotter::Utils.save_image(image, save_path, file_name)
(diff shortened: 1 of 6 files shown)


mlandauer (changes requested):
Thanks for cleaning up the git history, fixing that weird travis CI problem and getting this PR in.

There's quite a bit of feedback here. It's really lots of stylistic things and good practise - about writing code that is easy for other humans to read and understand. With a bit of luck it won't take you too long to work your way through.

Thanks!

lib/tasks/application.rake
```diff
@@ -168,4 +168,56 @@ namespace :application do
...
+      PolicyPersonDistance.find_each do |ppd|
+        temp = []
+        member = Person.find(ppd.person_id).latest_member
+        policy = Policy.find(ppd.policy_id)
+        temp << "http://#{ActionMailer::Base.default_url_options[:host]}#{member_policy_path_simple(member, policy)}?card=true&pp=disable"
+        temp << "#{member.id}_#{policy.id}.png"
+        urls_and_name << temp
+      end
+      save_path = Rails.root.join("app/assets/images/production_cards/member_policy_vote")
+      task("application:generate:capture_screenshots").invoke(urls_and_name, save_path, 600, 350)
```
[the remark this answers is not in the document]
Taz17: Width and height of the card image, they now are stored in variables.

Taz17:
@mlandauer this is ready for another review. Note, I am currently working on the naming of the images I intentionally pushed my commits so for feedback on the current changes

mlandauer:
@Taz17 a general comment about the way you're using git. Each commit needs to be an **atomic** change. It's good practise to make your changes small but for instance in 502baf8a031d2dc2a82cc86f1f621e5fe621c26b you're just deleting a whole bunch of code without explanation or context. Imagine yourself looking at that change and wondering why it was there. It's **so** much more useful if the change you're committing is the refactoring where you've moved code from the rake task to a new module. So then before the commit the code works and after that single commit the code works. Then also the refactoring is nicely captured in a single commit. And anyone reading that commit can understand exactly why it's there. Also if you need to undo a change for some reason it's super easy. You just do a "git revert".

mlandauer:
Also those commit messages are just awful. For instance, "remove rake task" - yes you removed a rake task. Technically it's accurate but it's not helpful. What rake task? Why? Don't change say **what** change you're making. Say **why** you're making it. Imagine yourself reading the commit message without any context. Does it make sense? Do you have any idea what it's talking about? Imagine scanning through a hundred commit messages trying to find the right one? Is this going to help?

mlandauer (commented):
Looking better. I think this is good enough to merge once you address the naming of the file. Thanks @Taz17

Taz17:
@mlandauer I have committed the changes regarding the naming of the files. I simply replaced replaced `/` with `_` in the path to the URL. This would be both to easy to read and easy to access within the application.

app/lib/card_screenshotter/members.rb
```diff
@@ -0,0 +1,27 @@
...
+
+      card_width = 600
+      card_height = 350
+      save_path = Rails.root.join("public/cards/member_policy_vote")
+
+      PolicyPersonDistance.find_each do |ppd|
+        person = ppd.person
+        policy = ppd.policy
+        url = "http://#{ActionMailer::Base.default_url_options[:host]}#{person_policy_path_simple(person, policy)}?card=true"
+        file_name = person_policy_path_simple(person, policy).gsub("/", "_")
```
mlandauer, line 18: What's an example url for one of the generated images and the corresponding url for the person policy page that it's related to?

Taz17: URL: `https://theyvoteforyou.org.au/people/senate/victoria/raff_ciccone/policies/86`
file name: `_people_senate_victoria_raff_ciccone_policies_86.png`

mlandauer: I'm specifically asking for the **url** that you would access the generated image from.

Taz17: I don't know, I have not been able to serve images that are in `public` directory to see how the URL would look like. How would I do that, I cannot find a resource to link images outside of the asset pipeline

mlandauer: Rails should just serve stuff from `public` in development. In production `public` will get served by nginx. So, for example if you go to http://localhost:3000/robots.txt it will serve the contents of `public/robots.txt`. Try it.

Taz17: Oh, the URL would look like this `http://localhost:3000/cards/member_policy_vote/_people_senate_act_kate_lundy_policies_1.png`

mlandauer: Thanks. In that case what would it take for the URL for the generated image to be?
https://theyvoteforyou.org.au/cards/people/senate/victoria/raff_ciccone/policies/86.png

Taz17: Since you cannot have the `/` character in a file name you would have to create the directory `public/cards/people/senate/victoria/raff_ciccone/policies/` so that the URL can have the `/cards/people/senate/victoria/raff_ciccone/policies/` and the image name would `86.png`

Taz17: I have made that change, i'll push it now.

mlandauer:
Merged. Thanks @Taz17

PR #1367, opened 2022-03-08
Socialcard2

tseyiokorodudu:
This PR contains the code responsible for the view side of social card 2. It includes a `member_card.html.haml` file, an update to the show action on the `members_controller` and `card.scss` file.


mlandauer:
Thanks! That's also now rolled to production

PR #1373, opened 2022-03-15
Let the width and height be determined by the creator of the screenshotter

Taz17:
These changes are made since not all cards are of the same size. Hence, let the caller determine the size they need upon initialising a new instance


mlandauer:
Closing this (and #1374) as we will be using the same image size for all the social sharing cards

Issue #1369, opened 2022-03-08
Social Card 3: Sharing a Policy Page


tseyiokorodudu:
### Social Card 3 is for when someone shares a policy page [https://theyvoteforyou.org.au/policies/86](https://theyvoteforyou.org.au/policies/86)

Which currently looks like this when it is shared. 

![](https://user-images.githubusercontent.com/31440427/153291685-cb6e00af-2ae3-49db-b04b-6449b65fb7e2.png)

The initial design of the card as displayed in issue #1316 is this 

![](https://user-images.githubusercontent.com/31440427/153294141-29854361-4d98-40d2-9662-941c9a10c9ff.png)

tseyiokorodudu:
## Current Design

Based on @mlandauer comment https://github.com/openaustralia/publicwhip/issues/1316#issuecomment-1036728795 and that of @katska https://github.com/openaustralia/publicwhip/issues/1316#issuecomment-1033337833 I have worked on the design further to create the image below.

![image-2](https://user-images.githubusercontent.com/31440427/157172463-4401198d-ef7a-4a56-b32a-c4b6bae75518.png)

Right now the member images are hardcoded (aka still working on it). However, the plan as previously stated in issue #1316 is ...

> The plan for the images on design 2 was to order and pick them randomly. However, @Taz17 thinks this might affect the consistency of the image people get when they share the card (i.e the MPs images on the card when person A shares it would differ from when person B shares it). So he suggested this ...

> Random but re-producible so that when two people share the same page they get the same image, hence one does not look more recent than the other.

Taz17:
For now, we'll be working on a method to randomly produce the images (but still be able to re-produce the same order). The policy number (which is unique) will be the seed for the randomness.

Taz17:
Here is what a card looks like when it the images are scrambled. 
![image](https://user-images.githubusercontent.com/86769131/157380042-a64f9696-8021-4c6e-9187-8475bcf15f7a.png)

tseyiokorodudu:
> Here is what a card looks like when it the images are scrambled. ![image](https://user-images.githubusercontent.com/86769131/157380042-a64f9696-8021-4c6e-9187-8475bcf15f7a.png)

@katska and @mlandauer what are your thoughts on this card design and implementation?

Things to note:
- The images displayed were picked using a method that does what @Taz17 describes below:

[...]

PR #1374, opened 2022-03-15
Now lets members determine the screenshotters width and height

Taz17:
Since not all cards are the same size. Member now has its own `CARD_WIDTH` and `CARD_HEIGHT` constants and are passed onto the screenshotted upon init

This PR can only be merged after https://github.com/openaustralia/publicwhip/pull/1373 is reviewed and merged. Changes in this PR depend on changes made in https://github.com/openaustralia/publicwhip/pull/1373.

### DIFF
--- a/app/lib/card_screenshotter/members.rb
+++ b/app/lib/card_screenshotter/members.rb
@@ -5,9 +5,11 @@ module CardScreenshotter
     class << self
       include Rails.application.routes.url_helpers
       include PathHelper
+      CARD_WIDTH = 600
+      CARD_HEIGHT = 350
 
       def update_screenshots
-        screenshotter = CardScreenshotter::Utils.new
+        screenshotter = CardScreenshotter::Utils.new(CARD_WIDTH, CARD_HEIGHT)
         ppds = PolicyPersonDistance.all
         progress = ProgressBar.create(title: "Members screenshots", total: ppds.count, format: "%t: |%B| %E %a")
         ppds.find_each do |ppd|


mlandauer:
I don't understand why the card size needs to change. I thought it was determined by the recommended sizes by Twitter and Facebook.

According to https://sproutsocial.com/insights/social-media-image-sizes-guide/#facebook the facebook recommended card size is 1200 x 628 and https://sproutsocial.com/insights/social-media-image-sizes-guide/#twitter says 1200 x 628 for Twitter as well.

So where did 600 x 350 in the current code come from?

mlandauer:
Looking in more detail at the Twitter documentation for the "Summary with large image" card https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image it says for the `twitter:image` attribute:
> A URL to a unique image representing the content of the page. You should not use a generic image such as your website logo, author photo, or other image that spans multiple pages. Images for this Card support an aspect ratio of 2:1 with minimum dimensions of 300x157 or maximum of 4096x4096 pixels. Images must be less than 5MB in size. JPG, PNG, WEBP and GIF formats are supported. Only the first frame of an animated GIF will be used. SVG is not supported.

For facebook in its "best practices" section https://developers.facebook.com/docs/sharing/best-practices#images:
> Use images that are at least 1080 pixels in width for best display on high resolution devices. At the minimum, you should use images that are 600 pixels in width to display image link ads. We recommend using 1:1 images in your ad creatives for better performance with image link ads.

So to me it does look like we should be using larger images than are currently being generated. For the card on righttoknow it appears to be 1200 x 630. See https://www.righttoknow.org.au/assets/logo-opengraph-a2b4e6d0abab52b292b71d8d829caebc16101924522f52ee0f84bb9af34aa2b0.png

mlandauer:
@Taz17 @tseyiokorodudu could you please both look at the image size question above as a matter of urgency because it pretty much effects everything else.

Taz17:
Why we chose 600 x 350:
@tseyiokorodudu and I chose 600 x 350 as it was within the range that twitter had specified (also works for facebook). It also was very close to the 2:1 ratio.

Why are the card sizes not all the same:
Different cards have different elements which need to be spaced out such that it looks neat. With the policies view card [see this image](https://github.com/openaustralia/publicwhip/issues/1369#issuecomment-1062575635) the height is 425 rather than 350 so that looks neater.

If we were to move to a 1200 x 630 size we would have to increase font, and image size. CSS lists will take care of the spacing/overflow.

mlandauer:
I'm going to do a test of sharing the current image size. I'll do it by using ngrok to expose my local dev server to the internet for the purposes of running it through the facebook & twitter testers.

mlandauer:
This is how the facebook sharing debugger is saying the current image size of 600 x 350 would be shown:

image

You can see the top and the bottom of the image are getting cropped.

tseyiokorodudu:
@mlandauer Thanks for testing it out. 

We have resized the cards, and their content to fit the recommended size of 1200 X 628. I just made a PR for that https://github.com/openaustralia/publicwhip/pull/1375

**Note:** The MP's image on the card now has the following dimensions `width: 337px;` `height: 450px;`.

mlandauer:
Closing this as we will be using the same size 1200 x 628 for all the social cards

Issue #1339, opened 2022-02-15
Social Card 1: When sharing a member's vote for a particular policy.


tseyiokorodudu:
### This issue is linked to issue #1316

This card is for when someone shares ["a member's vote on a particular policy" page](https://theyvoteforyou.org.au/people/senate/tasmania/eric_abetz/policies/162). 

This is the original design shared on issue #1316 
![](https://user-images.githubusercontent.com/31440427/153092295-799d20eb-f905-4f89-b384-a279c19db840.png)

Any further correspondence about this particular card would be done here.

tseyiokorodudu:
> For design 1, I opted for black bold text as opposed to white because I realised that white against the coloured background was a little hard to read and also had a low accessibility contrast. 

In issue #1316 I explained why we chose the text color for the initial card design. I've attached an image using the black text on the coloured background and one using the white text so we can see what I mean by my comment. 


![100001_1-2](https://user-images.githubusercontent.com/31440427/153995820-e0ef6296-c5ef-46da-a779-eb9572ca1616.png)


![100001_1](https://user-images.githubusercontent.com/31440427/153995840-3796ffff-6d20-4ae9-a684-28a0a73d3832.png)

mlandauer:
For anyone coming to this issue blind it's been split out from #1316. This issue just covers the design and implementation of one of the cards. #1316 covers several cards.

mlandauer:
@tseyiokorodudu do you want me to give feedback on the versions in https://github.com/openaustralia/publicwhip/issues/1339#issuecomment-1039871092 or were they purely meant to demonstrate the contrast issue? I'm asking because it's not clear to me and there have been some other changes made.

tseyiokorodudu:
@mlandauer Yeah I would like some feedback. Sorry for not making that clear.

mlandauer:
Here's some more feedback in no particular order again:

* Looks clearer with "They Vote For You" at the top
* I'm guessing from the text that this is actual rendered content from the implementation? If so, great!
* The font looks a bit weird and jagged to me. Is it the same font as used on the site?
* I agree that the contrast is way too low with the white writing. It's annoying because we have the same issue on the landing page of the site. At least on the site it's absolutely giant writing, which I guess helps to make it somewhat readable even if it isn't sticking to any accessibility standard.
[...]

PR #1375, opened 2022-03-16
Social card 1 & 2 resize

tseyiokorodudu:
Resized the cards and their contents to fit the recommended dimension of 1200 X 628. 

Social Card 1, i.e. Policy with member card, now looks like this:
 
image

Social Card 2, i.e. Member card looks like this: 

image

### DIFF
--- a/app/assets/stylesheets/card/_card.scss
+++ b/app/assets/stylesheets/card/_card.scss
@@ -1,60 +1,59 @@
 .socialmedia-card-container {
-    font-family: 'Fira Sans', sans-serif;
-    width: 1200px;
-    height: 628px;
-    background-color: $color-lemon-light;
+  font-family: 'Fira Sans', sans-serif;
+  width: 1200px;
+  height: 628px;
+  background-color: $color-lemon-light;
 }
-  
+
 .socialmedia-card-logo {
-    font-size: 20px;
-    background-color: $color-lemon;
-    padding: 0.75em 3.25em;
+  font-size: 32px;
+  background-color: $color-lemon;
+  padding: 0.5em 1.75em;
 }
-  
+
 .socialmedia-card-image-card {
-    display: grid;
-    grid-template-columns: 30% 1fr;
-    grid-column-gap: 100px;
-    grid-template-areas: 'a b';
-    margin: auto;
-    width: 90%;
-    padding-top: 1.4em;
-    padding-bottom: 1.4em;
+  display: grid;
+  grid-template-columns: 30% 1fr;
+  grid-column-gap: 35px;
+  grid-template-areas: 'a b';
+  margin: auto;
+  width: 90%;
+  padding-top: 3em;
+  padding-bottom: 3em;
 }
 
 .socialmedia-card-member-image {
-    grid-area: a;
-    img {
-        width: 337px;
-        height: 450px;
-    }
+  grid-area: a;
+  img {
+    width: 280px;
+    height: 370px;
+  }
 }
 
 .fitin {
-    overflow: hidden;
-    max-height: 450px;
-    display: flex;
-    justify-content: center;
-    align-items: center;
+  overflow: hidden;
+  max-height: 370px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
 }
 
 .socialmedia-card-heading-text {
-    grid-area: b;
-    color: #262626;
-    font-size: 67px;
-    line-height: 1.3;
+  grid-area: b;
+  color: #262626;
+  font-size: 72px;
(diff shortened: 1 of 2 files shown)


mlandauer:
The font sizes are now completely different than they were before. I don't understand why for instance the size of the image has almost doubled but the "logo" has only gone from 18px to 20px, for example.

mlandauer:
Also please use the 2 space indenting as used everywhere else on the project

katska:
@mlandauer yep, those are imbalanced now 
@tseyiokorodudu @Taz17 it looks like all the elements and spacing aren't size responsive and are thrown out of whack now the image is bigger. Can you fix?

tseyiokorodudu:
So I doubled the font sizes and image size of the original value. 

image

image

mlandauer:
Comparing this to https://github.com/openaustralia/publicwhip/issues/1339#issuecomment-1046368699 I can see that the new version is much more cramped with less whitespace at the bottom and top.

katska:
@tseyiokorodudu Looks much better already! Keep going to match the new size to the design layout that was approved before! Here's the image from #1339 for easy reference 
154872371-54c63527-27ed-4b7b-a3f2-522b223efe0c

mlandauer:
Testing out the new size with Facebook gives this:

image

Looks good so far. The little badge in the top left hand corner is just because this is running from my local dev box. Next Twitter

mlandauer:
And this is how it looks on Twitter

image

Not perfect but perfectly workable IMHO.

@katska @tseyiokorodudu do you think this is all good enough or do you think we should tweak the header and footer a little to take account of the slight crop on Twitter?

tseyiokorodudu:
I think we should. Maybe move the logo down a bit and the link up a bit.

mlandauer:
> I think we should. Maybe move the logo down a bit and the link up a bit.

Okay cool. So keep in mind that the design needs to work when viewed at the original full resolution as well as when it's cropped to 2:1.

mlandauer:
Without taking a measuring device out I would guess that Twitter is cropping 14 pixels off the top and 14 pixels at the bottom to make an image of 1200 x 600 (2:1).

katska:
@tseyiokorodudu @mlandauer agreed those borders look worth fixing. 
Spacing around the profile images looks good. 
There are questions about artefacts, such as how well the profile images hold up to resizing, or additional challenges coming from image translation. 
Specifically at this stage, I have a question about colours, which seem to become cooler out the other side of twitter and Facebook screenshots Matthew shows above. Is there something in any of the image settings that might account for this? The yellows have lost their ping!

mlandauer:
> Specifically at this stage, I have a question about colours, which seem to become cooler out the other side of twitter and Facebook screenshots Matthew shows above. Is there something in any of the image settings that might account for this? The yellows have lost their ping!

It's perfectly possible that something has gone awry with the colours between the web browser, the screenshotting process and whatever facebook and twitter are doing. Also, it might be the case that once this thing is rolled out in production the colours are different again. In my opinion that's not on the critical path. I think that's something that's easier to work on and try to address when everything is rolled out in production. We can see what's really going on rather than what we think is going to happen now.

Also, the usability of the cards is not really affected by any subtle colour shifts.

mlandauer:
@tseyiokorodudu I'll try the latest version on Twitter & Facebook now

mlandauer:
Facebook:

image

Twitter:

image

mlandauer:
I'm a bit confused because the images on the twitter and facebook tester look the same to me before and after. Maybe they're being cached or I made a mistake?

mlandauer:
Doing it all again with a different person/policy so I can be sure that there isn't some weird caching issue.

Facebook:
image

Twitter:
image

That looks better to me. I'm happy. Is this good enough for you @katska @tseyiokorodudu?

tseyiokorodudu:
Yeah, it is good enough.

mlandauer:
I'm going to merge this and roll it out to production. Before I run it in production I'll need to increase the size of the attached disk storage because it looks like image sizes have gone up from around 100kb to 250kb.

Based on the estimate of disk storage in https://github.com/openaustralia/publicwhip/issues/1338#issuecomment-1062439978 which now increases to 2.5 x 6 = 15 GB. So, with a little room to grow let's allocate 30 GB instead of the 10 GB we've got allocated at the moment.

mlandauer:
I'm running the screenshotting in production now. It says it's going to take over 3 hours all together. There's about 2:30 hours left. @tseyiokorodudu I can do the update to the front end for sharing of social card 1 when that's finished. Probably best to focus on the other cards now

Issue #1388, opened 2022-04-05
Card titles don't always shrink to fit in the card.


Taz17:
I visited https://theyvoteforyou.org.au/policies/86?card=true earlier and saw that the card title had not been adjusted to fit the card.

Here is a screenshot.

![image](https://user-images.githubusercontent.com/86769131/161658801-a5f1f229-064f-4604-b3ad-18790eb47e74.png)

This is hard to reproduce since it is caused by the event listener firing off before the html is actually rendered.

I did some digging to find out that `DOMContentLoaded` from the [MDN web docs](https://developer.mozilla.org/en-US/docs/web/api/window/domcontentloaded_event):

> The DOMContentLoaded event fires when the initial HTML document has been completely loaded and parsed, without waiting for stylesheets, images, and subframes to finish loading.

And the `load` event also from [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event): 

> The load event is fired when the whole page has loaded, including all dependent resources such as stylesheets and images. This is in contrast to [DOMContentLoaded](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event), which is fired as soon as the page DOM has been loaded, without waiting for resources to finish loading.

A fix to this issue is changing the event listener to `load` rather than `DOMContentLoaded`

Taz17:
Closing is the PR has been merged

PR #1389, opened 2022-04-05
Changed `DOMContentLoaded` to `load` in all eventlisteners for all social media cards

Taz17:
Fixes issue #1388

Changes all event listeners to fire off on `load` rather than on `DOMContentLoaded`

### DIFF
--- a/app/views/card/member_card.html.haml
+++ b/app/views/card/member_card.html.haml
@@ -16,7 +16,7 @@
       = remove_slash(get_hostname(root_url) + member_path_simple(@member)) 
 
 :javascript
-  window.addEventListener('DOMContentLoaded', (event) => {
+  window.addEventListener('load', (event) => {
     fitin = document.querySelector('.fitin');
     fitin_inner = document.querySelector('div .socialmedia-card-heading-text');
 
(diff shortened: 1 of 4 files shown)


mlandauer:
@Taz17 have you tested how this change interacts with the screenshot code? My first thought is if the javascript runs later is there now a possibility that the screenshot will get taken before the javascript has run?

I guess it's a question of testing and understanding at what point exactly the screenshot happens.

Taz17:
I am trying to test this out... Google chrome updated to version 100.xx and now the gem is not working. Im trying to downgrade my chrome verison right now.

mlandauer:
> I am trying to test this out... Google chrome updated to version 100.xx and now the gem is not working. Im trying to fix this issue now.

I don't think you need to downgrade google chrome. You can upgrade the chrome driver instead. That should be simpler.

Taz17:
@mlandauer The screenshots come out just fine as they should. The JavaScript is being executed before the screenshot is taken.

The difference between the `DOMContentLoaded` and `load` event listeners is that `DOMContentLoaded` does not wait for images or stylesheets. So the javascript executes before everything is positioned and **font size is set** hence causing the card to render the way it did as the screenshot I posted in the issue.

Using the `load` event listener means the text will shrink once styling and images are rendered. Hence avoiding that render I posted

PR #1391, opened 2022-04-05
Social card 4 revisions

Taz17:
This PR adds changes to social card 4 after receiving feedback.

What changed:

- Space around member image is now the same all around the image
- Number of policies left does not appear if there are zero left.
- Policies' font size on the card shrink when the last list item is lower than the URL on the card.

### DIFF
--- a/app/assets/stylesheets/card/_card.scss
+++ b/app/assets/stylesheets/card/_card.scss
@@ -87,34 +87,46 @@
 .socialmedia-card-member-votes-element {
   margin: auto;
   width: 90%;
-  padding-top: 1em;
-  padding-bottom: 1.5em;
-}
-.socialmedia-card-member-votes-text {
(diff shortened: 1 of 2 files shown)


mlandauer:
@Taz17 can you please add a screenshot of the before and after of this change. Thank you!

Taz17:
Sure!

Here is the card before any changes made:

Notes:

- **The slight difference between the colours in both images is due to the fact that one is a screenshot from a windows machine using chrome and the other is from a linux machine using firefox**
- I also forgot to mention that we are using bullet points to display the list of policies and that we fixed the indentation of the list.
- In the 'After' image the text has been shrunk such that the policies do not overflow to the linked URL.

Before

image


After

image


Here is a screenshot before the text on the card is shrunk.

image

PR #1392, opened 2022-04-06
Fix spelling

cofiem:


mlandauer:
Thank you @cofiem for spotting this mistake and fixing it!

PR #1525, opened 2025-08-12
Remove t shirt banner that we ran across 2025 federal election

mlandauer:
We've now run out of XL and L sizes. Only medium and small left and sales have tapered off so makes sense to close off.


simonzippy (approved):
Looks good to go. Just noting for future reference we still have the view in place for the swag T-shirt here [app/views/application/_swag.html.haml](https://github.com/openaustralia/theyvoteforyou/blob/5873344ebbf2bee4fa10e863bfff6159f508fd2c/app/views/application/_swag.html.haml) but the reference to use it has been commented out.

PR #1536, opened 2025-10-06
set deploy branch to main

Br3nda:
also, staging was using a branch named "test". Took me a while to work out why deploying to staging wasn't changing staging.

### DIFF
--- a/config/deploy/staging.rb
+++ b/config/deploy/staging.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-set :branch, :test
+set :branch, :main
 set :deploy_to, "/srv/www/staging"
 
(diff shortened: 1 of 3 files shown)


config/deploy/staging.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-set :branch, :test
+set :branch, :main
```
jamezpolley, line 3: I think that having this different should signify that changes are being tested on the test then merged into main. Maybe that hasn't often happened, but to me it feels more correct to leave this at test.

From your comments @Br3nda it sounds like you're used to something different though. I don't think it matters a huge amount anyway as this can be set manually any time someone is deploying if they want to use a different branch.

I'm going to go ahead and approve this - easy to come back and tweak later if a need arises.

PR #1547, opened 2025-10-07
Adds continuous deployment to staging

Br3nda:
https://github.com/marketplace/actions/capistrano-deploy


Br3nda:
not happy yet with how secure this is. I'm gonna close this and look for another approach

PR #1579, opened 2026-03-11
Add github-actions ecosystem to Dependabot

jbampton:
https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#package-ecosystem-

Group updates to reduce noise

Add cooldown period

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,7 +14,7 @@ updates:
     schedule:
       interval: "weekly"
     groups:
-      github-dependencies:
+      github-actions-dependencies:
         patterns:
           - '*'
     cooldown:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,13 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "weekly"
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      github-dependencies:
+        patterns:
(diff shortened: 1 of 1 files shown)


ianheggie-oaf (commented):
If I understand correctly this:
* reduces inspection frequency to weekly - **I like this (less noise)**
* Groups the updates into one per week - This shifts the pain from too many issues (one per gem version update) to less updates, but title no longer gives what is being updated - just that it is "github-dependencies" **- I do NOT like this aspect, my guess is that it will be a bit more effort, not less**
* waits for 7 days after a gem is released (presumably in case a fix to the fix becomes quickly needed, but can block gem from updating if there are a number of releases quickly in a row) - **Overall I think this is a good idea (less noise, not as close to bleeding edge)**

Overall assessment: Meh - worth a try, interested in what @Br3nda and @benrfairless think

.github/dependabot.yml
```diff
@@ -9,3 +9,13 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "weekly"
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      github-dependencies:
```
jbampton:
```suggestion
      github-actions-dependencies:
```

jbampton (commented):
Minor update to the group label.

This group label shows in the dependabot PR titles.

Small improvement

benrfairless (approved):

PR #1586, opened 2026-03-17
misc: fix spelling

jbampton:
General clean up

### DIFF
--- a/app/lib/data_loader/division_xml.rb
+++ b/app/lib/data_loader/division_xml.rb
@@ -37,12 +37,12 @@ module DataLoader
 
     def debate_url
       # TODO: PHP always gets the previous heading, major or minor. Is this to support missing headings?
-      preceeding_minor_heading_element.attr(:url)
+      preceding_minor_heading_element.attr(:url)
     end
 
     def debate_gid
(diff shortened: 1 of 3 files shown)


ianheggie-oaf (approved):
I approve on the basis that
A. [Section headings are ignored](https://github.com/php/php-src/blob/38cbbee4aa14d2832686eeb771b9e8fa1539aa4e/php.ini-production#L24), which is a surprise to me, and so the change is cosmetic
B. All the calls to the two methods seems to have been corrected at the same time:
```
ianh@ben:.../theyvoteforyou (main)$ ag preceeding_minor_heading_element
app/lib/data_loader/division_xml.rb
40:      preceeding_minor_heading_element.attr(:url)
45:      preceeding_minor_heading_element.attr(:id)
96:    def preceeding_minor_heading_element
101:      preceeding_minor_heading_element.inner_text.strip
ianh@ben:.../theyvoteforyou (main)$ ag preceeding_major_heading_element
app/lib/data_loader/division_xml.rb
88:    def preceeding_major_heading_element
93:      preceeding_major_heading_element.inner_text.strip
```

PR #1616, opened 2026-05-21
Feature/tag deployments in git log

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/theyvoteforyou/issues/1612
* 

## What does this do?

Tags when production and staging are deployed for KPI stats

based on (review and merge first):
* https://github.com/openaustralia/theyvoteforyou/pull/1615

## Why was this needed?

One of my KPI is release cadence and timeliness

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -137,6 +137,7 @@ group :development do
   gem "capistrano", require: false
   gem "capistrano-rails", require: false
   gem "capistrano-rvm"
+  gem "capistrano-tagging3", "~> 2.0"
   gem "capistrano-maintenance", require: false
   gem "ed25519", "~> 1.3"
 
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -163,9 +164,9 @@ end
 group :doc do
   # bundle exec rake doc:rails generates the API under doc/api.
   gem "sdoc", require: false
-  # Fix Unresolved or ambiguous specs
-  gem "psych", "~> 5.2.6"
 end
+# Fix Unresolved or ambiguous specs
+gem "psych", "~> 5.2.6"
```
benrfairless, line 169: Are these commented correctly?

ianheggie-oaf: @benrfairless The comment explains why - I was literally getting a warning from bundler about "Unresolved or ambiguous specs" re psych - the gem line is my attempt to resolve. If it doesn't then I will have to look more carefully on the order of the gems included.

benrfairless (approved):

Gemfile
benrfairless: My comment here was about the indentation of the text - not about the gem itself. I'm happy for this to go through as long as that is OK :)

PR #1624, opened 2026-05-28
Only require capistrano gems when using capistrano

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/oaf-internal/issues/243

## What does this do?

Fixes tagging3 trying to pull in capistrano when its not deploying

## Why was this needed?

tests where breaking


ianheggie-oaf:
Rebased on main and the changes had been merged with another PR - so its closed

Issue #1656, opened 2026-07-21
Update TheyVoteForYou from MySql 8.0 to 8.4 before AWS charges in August 1, 2026


ianheggie-oaf:
**Problem**

Aws charges extra after [RDS end of standard support date](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/MySQL.Concepts.VersionMgmt.html) 

| MySQL engine version | Community release date | RDS release date | RDS end of standard support date |
| --- | --- | --- | --- |
| 8.0.45 | 20 January 2026 | 3 February 2026 | **31 July 2026** |
| 8.4.9 | 21 April 2026 | 8 May 2026 | 8 May 2027 |

**Describe the solution you'd like**

1. Update docker-stack (now devcontainer) so it works on a dev system that already has mysql (override mysql port), and is pinned to mysql version 8.4
  * https://github.com/openaustralia/theyvoteforyou/pull/1657
2. Adjust  config/database.yml so it can be configured to a different mysql port
  * https://github.com/openaustralia/theyvoteforyou/pull/1657 (same PR)
3. Make sure the software handles mysql 8.4 (run tests, static check of code) - yes
  * https://github.com/openaustralia/theyvoteforyou/pull/1658
4. Upgrade our software on production - not a blocker => defer
5. Upgrade the RDS database
  - [x] Update rds_admin_password so it uses the newer encryption standard (Turns out AWS handles that itself as part of upgrade? or uses its own plugin that uses the old standard?)
  - [x] Update staging passwords so it uses the newer encryption standard
  - [x] Update production passwords so it uses the newer encryption standard
  - [x] Upgrade the database engine itself (via terraform)

**Additional context**

See also issues:
* https://github.com/openaustralia/openaustralia/issues/757
* https://github.com/openaustralia/infrastructure/issues/597

I checked TVFY's Gemfile* and the ruby gems appears ready, but I haven't tested it for any sql syntax deprecated in 8.0 and removed in 8.4. We will need to get a docker mysql8.4 up and running and test it (or have you done this already?).

What will need to happen is that we will need to enable the mysql_native_password plugin when we migrate to 8.4 (otherwise logins, even admin will fail) , and then we can upgrade all the passwords to the latest type before mysql 9.0 removes that plugin completely.

ianheggie-oaf:
## MySQL 8.0 → 8.4 RDS Upgrade — Compatibility Checklist

### Get local devcontainer working so I can test against different mysql versions

[...]

Issue #10, opened 2013-12-29
Add support for "pairs" in division data (PW-7)


oajirabot:
**Issue details imported from Jira:**

| Detail | Value |
| --- | --- |
| Assignee | Henare Degan |
| Reporter | Matthew Landauer |
| Created | 2009-03-31 22:11:39 |
| Updated | 2013-08-20 20:27:46 |

---
### Imported Jira Comments

**Matthew Landauer** - 2009-09-01 11:17:32

> <p>I'm pretty sure I've updated the parser to collect and output the pairs data. Need to check and if so close this ticket.</p>

**Henare Degan** - 2013-08-20 20:27:46

> <p>It's being parsed, here's a recent example: <a href="https://gist.github.com/henare/5480108#file-oa_2013-03-21-xml-L172" class="external-link">https://gist.github.com/henare/5480108#file-oa_2013-03-21-xml-L172</a></p>
>

PR #1657, opened 2026-07-22
Provide full devcontainer environment

ianheggie-oaf:
## Description

Provide a full devcontainer environment, including web container for the app, merging ideas from docker-stack directory by Brenda and local_dev from Matthew and others (mix of vagrant, docker and ansible) and planningalerts.

It is documented in the README.md and the out of date documents have been removed.

Helper make targets are included if you want to use the devcontainer CLI.

Stack provides:
* https://github.com/openaustralia/theyvoteforyou/issues/1667

Rolled PRs into this PR and closed them:
* https://github.com/openaustralia/theyvoteforyou/pull/1658 (we are on mysql 8.4 now anyway)
* https://github.com/openaustralia/theyvoteforyou/pull/1671 (start off the way we mean to continue)

Added two PRs before this one in the stack:
* https://github.com/openaustralia/theyvoteforyou/pull/1678
* https://github.com/openaustralia/theyvoteforyou/pull/1675

## Motivation and Context

This change was needed because there was not a clean development environment to do the work for:
* https://github.com/openaustralia/theyvoteforyou/issues/1656

Also satisfies (for this repo):
* https://github.com/openaustralia/theyvoteforyou/issues/1649
* https://github.com/openaustralia/oaf-internal/issues/307

This should be compatible with macOS (arm) systems, but needs to be tested on such a machine.

## How Has This Been Tested?

I have run the devcontainer using the CLI, checks I can create the database, run tests and run the web application.

Details of your testing environment:
Developed and tested on Ubuntu 24.04.4 LTS x86_64 with ruby installed by mise, mysql 8.0, redis 5:7.0, make 4.3
```
$ bin/setup-host 
OK: Docker + Compose plugin already installed
OK: devcontainer CLI already installed

== Versions ==
Docker version 29.6.2, build dfc4efb
Docker Compose version v5.3.1
Devcontainer: 0.72.0
```
- [x] Ran automated tests on my own system
- [ ] Confirmed it passed the GitHub actions tests

### Resource Usage

Elasticsearch was running with no explicit heap size, causing it to auto-size
based on detected host memory (48GB) and used ~25GB even for a small dev index. I've
set explicit `mem_limit` and heap/buffer tuning on each service (elasticsearch,
mysql, rails-app, mailpit) so the whole stack fits within a ~6GB budget, which
should be usable on a 16GB dev machine without swapping.

Verified this doesn't affect functionality or performance: ran `bin/rake spec`
three times before and after the change (28.4s / 28.9s / 28.8s, no regression),
and confirmed via `docker stats` during and after a test run that every service
stays well within its new limit with headroom to spare (e.g. elasticsearch
peaked at 1.3GB against its 3GB cap).

## Types of Changes

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change) - removed Vagrant and old docker setup
- [x] Documentation

## Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.


benrfairless (commented):
Thanks Ian, this is a solid piece of work and the direction is right. Deleting 644 lines of three overlapping half-maintained environments is the sort of cleanup that usually never happens, and `bin/setup-host` is genuinely careful.

**I've tested it on macOS 27 / Apple silicon, and it works.** That closes the "VsCode to be tested by Ben on MacOS" item on #1667, at least for the CLI path. Details below.

I've found two things I'd like fixed before merge, plus a handful of smaller ones. Nothing here is structural, it's mostly config and docs that have drifted from what the code actually does.

## macOS arm64 test results

Tested on macOS 27.0, Apple silicon, 48 GB RAM, Docker 29.6.2 with an 8 GB VM, Compose v5.3.1, devcontainer CLI 0.88.0.

| Check | Result |
|---|---|
| `make dev-up` | success, all four services healthy |
| Architecture | `aarch64` native in rails-app and mysql, no QEMU emulation |
| Ruby | 3.4.4 `[aarch64-linux]`, matches `.ruby-version` |
| Elasticsearch | 7.17.7 reachable on 9288 |
| Mailpit | HTTP 200 on 8088 |
| Chromium | 150.0.7871.181 plus chromedriver, `CHROME_NO_SANDBOX=1` |
| `make dev-rake` | **374 examples, 0 failures**, 16.78s, coverage 65.61% |
| `make dev-server` | 200 on `http://localhost:3088/` with real content, `/admin/` 302 as expected |
| Mail pipeline | test email delivered via ActionMailer into Mailpit, confirmed through its API |
| OOM kills | none |

Same suite figures as your Linux run (374 examples, 0 failures, 65.61%), and faster on Apple silicon: 16.78s against your 24.23s.

Memory at idle sits well inside your 6 GB budget: elasticsearch 1.35 GiB of 3 GiB, mysql 276 MiB of 1.465 GiB, rails-app 403 MiB of 1 GiB, mailpit 18 MiB of 128 MiB. Your resource work checks out.

## Must fix

### 1. `config/database.yml.example` can silently connect to the developer's host MySQL

`config/database.yml.example:23-29`

```erb
<% if ENV["DB_HOST"] || !File.socket?("/var/run/mysqld/mysqld.sock") %>
  host: <%= ENV.fetch("DB_HOST") { "mysql" } %>
  port: <%= ENV.fetch("DB_PORT") { "3306" } %>
<% else %>
  socket: /var/run/mysqld/mysqld.sock
<% end %>
```

I rendered the template across the realistic cases rather than eyeballing it:

| Scenario | Resolves to |
|---|---|
| Inside the devcontainer (`DB_HOST=mysql`) | `host: mysql  port: 3306` ✅ |
| Host, following the README exactly | `host: 127.0.0.1  port: 3388` ✅ |
| **Host, local MySQL socket present, `DB_HOST` unset** | **`socket: /var/run/mysqld/mysqld.sock`** ❌ |
| **Host, `DB_HOST` set but `DB_PORT` forgotten** | **`host: 127.0.0.1  port: 3306`** ❌ |

The last two connect to the developer's own MySQL rather than the container. On your setup (Ubuntu 24.04 with mysql 8.0) that means `bin/setup` runs `db:prepare` and creates `tvfy-development` and `tvfy-test` on your host 8.0 server, and then the suite runs against **8.0** while appearing to test 8.4. Whether it fails loudly depends only on whether the host root password happens to be `password`.

macOS is accidentally safe here, since Homebrew's socket is at `/tmp/mysql.sock`, so it takes the host branch and fails loudly. That asymmetry means this will not show up in my testing, only on yours.

This is the collision #1656 item 1 exists to prevent, so I think it's worth getting right.

Suggested fix: drop the socket branch. Nothing in this stack ever has a local MySQL socket, so it only fires in the failure case.

### 2. `Gemfile.lock` is missing `aarch64-linux`, so every Apple silicon developer gets a dirty tree

Starting the devcontainer modified a tracked file on a clean checkout:

```diff
 PLATFORMS
+  aarch64-linux
   arm64-darwin-25
   x86_64-darwin-20
   x86_64-linux
```

plus `ffi (1.17.4-aarch64-linux-gnu)`, `libv8-node (18.19.0.0-aarch64-linux)` and `nokogiri (1.19.4-aarch64-linux-gnu)`.

The `bundler-cache` feature runs `bundle install` in the container, and Bundler adds the running platform. `x86_64-linux` is already listed, which is why you never saw this, but the devcontainer on an M-series Mac is `aarch64-linux`. So every macOS developer's first `make dev-up` leaves `Gemfile.lock` modified, and they either commit unrelated churn or work with a permanently dirty tree. It would also break a frozen or deployment-mode `bundle install`.

Since this PR makes arm64 Linux a supported development platform, I think the lockfile should say so:

```bash
bundle lock --add-platform aarch64-linux
```

## Should fix

<details>
<summary><b>3. <code>make dev-up</code> does not pick up <code>compose.yaml</code> changes</b></summary>

Found while testing the stack's top branch. `devcontainer up` passes `--no-recreate`:

```
docker compose --project-name theyvoteforyou -f .devcontainer/compose.yaml \
  -f ...devcontainer.build-....yml -f ...devcontainer.containerFeatures-....yml \
  up -d --no-recreate
```

So an existing container is reused even when its image has changed underneath it. On the #1658 branch, `make dev-up` reported `{"outcome":"success"}` and MySQL was still 8.0.46. `make dev-down && make dev-up` applies it correctly, and `dev-down` works cleanly despite passing a different `-f` set.

Worth a README line, or a target:

```make
dev-recreate:
	$(MAKE) dev-down
	$(MAKE) dev-up
```

</details>

<details>
<summary><b>4. The two SSL settings do not do what the comment says, and warn on every connection</b></summary>

`config/database.yml.example:32-34`, from a8a0b4ee7:

```yaml
# Ignore TLS/SSL error: self-signed certificate in certificate chain
ssl_mode: preferred
sslverify: false
```

Measured inside the container:

```
client_lib={id: 30409, version: "3.4.9", header_version: "3.4.9"}
SSL_MODE_PREFERRED=0
tls_cipher="TLS_AES_256_GCM_SHA384"
```

Id 30409 falls in the `[30000, 40000)` range that mysql2's own source comments label MariaDB Connector/C 3.x, which is what the `mysql-client` feature supplies. That connector does not define the `SSL_MODE_*` constants, so mysql2 falls back to `SSL_MODE_PREFERRED = 0` at build time (`ext/mysql2/client.c:1834`). Zero is neither `DISABLED` nor `REQUIRED`, so the setter warns and returns without calling `mysql_options` at all (`client.c:163-170`). That's the `does not support ssl_mode 0` warning visible in your own test output on #1658.

Meanwhile the connection is running TLS 1.3 regardless. `sslverify: false` does not set `SSL_VERIFY_SERVER_CERT`, which is off by default anyway; its only other effect is causing `ssl_set` to be called with all-nil arguments.

So the link is encrypted either way and these two lines look like they're contributing nothing but noise. What was the original error? Worth checking it stays fixed with both removed, because something else is doing the work.

</details>

<details>
<summary><b>5. <code>bin/docker-entrypoint</code> is dead code</b></summary>

Added in 25313b783. `grep -rn "docker-entrypoint"` across the repo returns nothing: no `ENTRYPOINT` in the Dockerfile, no `entrypoint:` in `compose.yaml`, and the service uses `command: sleep infinity`. It also runs `./bin/rails db:prepare` when it detects a server invocation, which would be surprising if it were ever wired up. Delete, or wire up deliberately.

</details>

<details>
<summary><b>6. <code>config/database.yml.example</code> comments describe things that don't exist</b></summary>

- `see mysql-init/ for where these dev/test users and databases get created` - there's no `mysql-init/` directory and nothing mounts one. The databases are created by `db:prepare` as root.
- `the "mysql" service name/port from docker-compose.yml, which is correct when Rails runs inside the "web" container` - the file is `.devcontainer/compose.yaml` and the service is `rails-app`.
- `export DB_PORT=13306` - the actual mapping is `3388`. The README gets this right, so the two contradict each other.
- The password-hashing link points at the MySQL **5.7** manual, in a stack about 8.4.

These read as carry-over from another project.

</details>

<details>
<summary><b>7. Dev email links point at a dead port</b></summary>

`config/environments/development.rb:41` still has `port: 3000` while the app is published on 3088. Confirmed on the running container:

```
mailer default_url_options={host: "localhost", port: 3000}
generated=http://localhost:3000/
smtp={address: "mailpit", port: 1025, ...}
```

The SMTP half is right and mail does reach Mailpit, it's only the links inside that are wrong. Since the README walks through checking signup emails, this will get hit early. Worth making env-driven alongside the Mailpit settings just above it.

</details>

<details>
<summary><b>8. The local compose override the <code>.gitignore</code> anticipates cannot work</b></summary>

eb9b1249b is titled "Update .gitignore so we can override docker-compose locally" and adds `/docker-compose.override.yml`. Two reasons it won't take effect: Compose looks for an override next to the compose file, so it'd need to be in `.devcontainer/`; and passing `-f` disables automatic override loading, which everything here does.

Since "override mysql port" is #1656 item 1, variable substitution would give a real knob with no override file:

```yaml
ports:
  - "127.0.0.1:${TVFY_MYSQL_PORT:-3388}:3306"
```

Exported in the shell before `make dev-up`, with the tracked default unchanged for everyone else.

</details>

<details>
<summary><b>9. <code>container_name: mailpit</code> will collide across projects</b></summary>

`.devcontainer/compose.yaml:84`. `container_name` is global to the Docker daemon, not scoped to the compose project, so no other stack using that literal name can run at the same time. #1667 says this is intended as a template for all projects, and Planning Alerts is the obvious candidate. Dropping the line gives `theyvoteforyou-mailpit-1`, which is unique and more discoverable.

</details>

<details>
<summary><b>10. <code>docker exec</code> cannot find Ruby, which I think is the RubyMine problem</b></summary>

The README asks why mise works via the devcontainer CLI but not in RubyMine. I think this is it:

```
$ docker exec theyvoteforyou-rails-app-1 sh -lc 'echo $PATH; ruby -v'
/home/vscode/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
env: 'ruby': No such file or directory

$ devcontainer exec --workspace-folder . bash -lc 'ruby -v'
ruby 3.4.4 (2025-05-14 revision a38531fd3f) +PRISM [aarch64-linux]
```

Neither the mise install dir nor `/home/vscode/.local/share/mise/shims` (which exists) is on `PATH`. `devcontainer exec` works because the CLI probes the user environment and injects it. RubyMine's Docker integration does a plain `docker exec` and gets the container's own `PATH`.

Probably fixable in the image so it holds for every entry path:

```dockerfile
ENV PATH=/home/vscode/.local/share/mise/shims:$PATH
```

Worth confirming against RubyMine before closing that item on #1667.

</details>

<details>
<summary><b>11. <code>.devcontainer/devcontainer-lock.json</code> is neither committed nor ignored</b></summary>

The CLI generates it on first `up`, pinning each feature by sha256 digest. Right now it's untracked and not gitignored, so it shows as stray output for everyone.

I'd commit it. The point of the stack is a reproducible environment, and the four features are currently floating on mutable tags. Related: `mysql-client` is requested with no version tag while the other three are pinned to a major, and per finding 4 that feature is what supplies the connector causing the ssl_mode warning.

</details>

<details>
<summary><b>12. The README's Elasticsearch health check will never show what it says</b></summary>

The README says look for `"status":"green"`. Measured here:

```json
{"status":"yellow","number_of_nodes":1,"active_primary_shards":6,"active_shards":6}
```

Yellow is correct and permanent for a single-node cluster holding indices with a replica setting above zero: primaries are all assigned, replicas can't be. It will never go green, so the first person to check will think their index is broken. Suggest saying yellow is expected on a single node, or pointing at `_cat/indices?v`, which is the more useful check anyway.

</details>

<details>
<summary><b>13. Other README bits</b></summary>

- Admin panel and feature flags sections still say `http://localhost:3000/admin/`, should be 3088.
- "add these to you `.envrc` file" - "your".
- The first-person note about RubyMine and mise reads oddly in a project README, and says "ruby-app container" where the service is `rails-app`. Maybe move to #1667 as a known issue.
- Getting started says to run `make dev-exec COMMAND="bin/setup --skip-server"`, but `devcontainer.json` already runs exactly that as `postCreateCommand`. Harmless since it's idempotent, but it reads as required.
- `brew install elasticsearch` no longer works, Elastic was removed from Homebrew core over the licence change.
- `## Using just the other services ` has trailing whitespace.

</details>

## Minor

- `.PHONY` omits `dev-down` and `dev-server`.
- `app/lib/card_screenshotter/utils.rb:27` uses `if ENV["CHROME_NO_SANDBOX"]`, which is truthy for `"0"` and for the empty string. `== "1"` would be tighter. Given it disables the Chromium sandbox, gating on `Rails.env.local?` too would mean a stray env var can't switch it on in production. Low real risk, since the screenshotter only renders our own pages.
- `.dockerignore` carries Kamal entries (`/config/deploy*.yml`, `/.kamal`) from the Rails template; we deploy with Capistrano. Just noise.
- `mysql` and `elasticsearch` declare `networks: [default]`, `rails-app` and `mailpit` don't. No functional difference, worth making uniform.
- Only one of the 29 commits carries an `Assisted-by:` trailer, and this PR's body doesn't mention AI assistance while #1658's does. Worth a pass for consistency with our convention.

## What I particularly liked

- `bin/setup-host` is careful: distro guard with a clear error, idempotent `docker_ready` and `devcontainer_ready` checks, `set -euo pipefail`, a bounded wait loop for Docker Desktop rather than a bare sleep, and the `vm.max_map_count` handling with a persisted sysctl. The comment explaining *why* Elasticsearch needs it, and that Docker Desktop already sets it, is exactly the comment that saves someone an afternoon.
- Health checks on all three dependencies with `condition: service_healthy`. That's what stops `postCreateCommand` racing MySQL, and it held up under testing.
- The memory work is properly evidenced: explicit limits per service, backed by three before-and-after timings and `docker stats` headroom. I re-measured and it holds.
- Elasticsearch pinned to match production with the verification method recorded in a comment.
- Ports bound to `127.0.0.1` rather than `0.0.0.0`, so a dev MySQL with a known root password isn't reachable from the network.
- Reflowing the README in its own commit (296df21b5) before changing content, which kept the real diff readable.
- Omitting Redis is correct rather than an oversight, since `config/cable.yml` uses `adapter: async` in development. I checked because its absence looked like a gap.

## Testing note

No test changes, and I don't think any are needed: the single application code change is one env-gated dev-only line, and `spec/lib/card_screenshotter/` was already empty before this stack. The real coverage gap is CI's MySQL version, which I've raised on #1658 instead.

---

*AI use disclosure: I put this review together with Claude (claude-opus-5) in Claude Code. The measured results quoted above come from actually running the devcontainer on my machine, and the mysql2 analysis was traced through the gem's source rather than recalled. I've read and I'm standing behind all of it, but do push back on anything that looks off.*

benrfairless (changes requested):

ianheggie-oaf:
RE:
>  1. config/database.yml.example` can silently connect to the developer's host MySQL

I feel on basis of
1. least surprise - since rails new on 8.0 creates a database.yml file with `socket: /var/run/mysqld/mysqld.sock` as the default, I feel it is reasonable for the example to fall back to that
2. I want the option of using my local host mysql service
3. The devcontainer sets DB_HOST so using the mysql container IS the default when using devcontainer (which I feel is a reasonable default within a devcontainer)

Thus I: Updated example database.env to detect socket path on macOS and use TCP connection if DB_PORT is set not just DB_HOST

> 2. Gemfile.lock` is missing `aarch64-linux`, so every Apple silicon developer gets a dirty tree

Resolved in 
* https://github.com/openaustralia/theyvoteforyou/pull/1671

> 3. make dev-up does not pick up compose.yaml changes

Resolved by making dev-up check if there where any relevant changes and running dev-down first if so, avoids `dev-recreate` target that you have to remember to run.

> 4. The two SSL settings do not do what the comment says, and warn on every connection

**Deferred till later**

> 5. bin/docker-entrypoint` is dead code

Removed.

> 6. config/database.yml.example comments describe things that don't exist

Updated example appropriately 

> 7. Dev email links point at a dead port

Server now runs with the same port as it is exported (3088)

> 8. The local compose override the .gitignore anticipates cannot work

removed .gitignore entry - did not add override ENV variables as I (hope) they wont be needed and thus wont add yet more README entries to read.

> 9. container_name: mailpit will collide across projects

Removed `container_name: mailpit` so the name is not set globally

> 10. docker exec cannot find Ruby, which I think is the RubyMine problem

**Deferred till later**

> 11. .devcontainer/devcontainer-lock.json is neither committed nor ignored

Investigated versions ... I had:
```
$ devcontainer --version
0.72.0
```
Vs Ben's
```
devcontainer CLI 0.88.0.
```

ianheggie-oaf:
@benrfairless 

● RE (remaining points):

> 4. The two SSL settings do not do what the comment says, and warn on every connection

Removed both `ssl_mode`/`sslverify` lines entirely rather than leaving them commented out. Ran the suite without them: passes fine, they weren't doing anything.

> 10. `docker exec` cannot find Ruby, which I think is the RubyMine problem

Still outstanding, this is the one remaining item from the whole review. Tracking it on #1667.

> 12. The README's Elasticsearch health check will never show what it says

Fixed properly rather than just reworded: dev/test indices now get `number_of_replicas: 0` (`config/initializers/elasticsearch.rb`), since a single-node dev/test cluster can never satisfy a replica anyway. Cluster health is genuinely green now, not
just documented as "yellow is fine".

> 13. Other README bits

All done:
* Admin panel / feature flags URLs: `:3000` → `:3088`
* "add these to you `.envrc`" → "your"
* RubyMine aside reworded to third person, `ruby-app` → `rails-app`, points at #1667
* `bin/setup --skip-server` step now notes it already ran via `postCreateCommand`
* `brew install elasticsearch` replaced (Homebrew dropped it after Elastic's licence change; the devcontainer/hybrid setup already provides Elasticsearch anyway)
* Trailing whitespace on the "Using just the other services" heading

**Minor:**
* `.PHONY` already had `dev-down`/`dev-server`, turned out not to be missing them
* `CHROME_NO_SANDBOX`: replaced the bare truthy check with `ENV.true?` (new helper backed by `ActiveModel::Type::Boolean`, `config/initializers/00_env_boolean.rb`), so `0`/`false`/`off` are false everywhere, not just here, with a spec
(`spec/lib/env_boolean_spec.rb`)
* `.dockerignore`: dropped the unused Kamal entries
* `mysql-client` feature pinned to major version 1, matching the other three and the lock file's resolved version
* `networks: [default]` added to `rails-app`/`mailpit` so all four services declare it consistently
* Assisted-by trailer consistency isn't retroactively fixable without rewriting shared history; adopting it going forward instead, now a documented convention in `CLAUDE.md`

**A few extras beyond the review:**
* `make dev-status`: container status, resource usage, and clickable service URLs
* Action Mailer previews wired up at `/rails/mailers`
* Previous PR in stack
  * `docs/DECISIONS.md`: a lightweight log for cross-cutting decisions that don't belong in one file's comments
  * `CLAUDE.md`: guidance for AI coding tools working in this repo


Only outstanding item now: **#10, the RubyMine/mise PATH fix**, still to do. (unfortunately its got worse, the containers run but it cannot connect - i will time box it because devcontainer CLI works, as does hybrid mode)

ianheggie-oaf:
Got it working with RubyMine! YIPEE!

.devcontainer/compose.yaml
```diff
@@ -0,0 +1,111 @@
...
+      mysql:
+        condition: service_healthy
+      elasticsearch:
+        condition: service_healthy
+      mailpit:
+        condition: service_healthy
+    networks:
+    - default
+  mysql:
+    image: mysql:8.4
```
benrfairless, line 39: **Question: which MySQL does production actually run?** Four sources disagree and they can't all be right:

- `.github/workflows/rubyonrails.yml` pins `mysql:5.7` with the comment "We're using mysql 5.7 in production currently"
- this line runs `mysql:8.4`
- the PR body closes #1658 on the grounds that "we are on mysql 8.4 now anyway"
- `config/database.yml.example` now documents `caching_sha2_password` as required, which is the 8.4 default

I can't verify production from here, but either answer has a consequence worth acting on in this PR, since this is the PR that introduces the gap:

- if prod is **8.4**, CI is testing against a version nobody runs and the workflow image + comment need updating
- if prod is **5.7**, dev now sits two majors ahead of both CI and prod, and auth/`db:prepare` behaviour can differ

Makefile
```diff
@@ -1,16 +1,127 @@
...
+
+LOCKFILE := .devcontainer/devcontainer-lock.json
+
+# Keeps the committed lockfile in sync with devcontainer.json (see finding 11
+# on #1656). `devcontainer upgrade` is a dedicated subcommand present in every
+# CLI version checked so far (identical output confirmed on both 0.72.0 and
+# 0.83.3), so this works regardless of whether a given install auto-generates
+# a lockfile from `up`/`build` by default.
+$(LOCKFILE): .devcontainer/devcontainer.json
+	devcontainer upgrade --workspace-folder .
```
benrfairless, line 55: **Question: should this be `upgrade`, or a re-pin?** `devcontainer upgrade` resolves features to *latest*, so the lockfile isn't really pinning anything: two developers editing `devcontainer.json` a month apart get different feature digests, and `make dev-up` silently dirties a tracked file in their working copy.

The comment cites "finding 11 on #1656" so there may be a rationale I can't see - is auto-upgrading the intent, or should this re-pin the versions already in the lockfile and leave upgrades as a deliberate act?

Makefile
```diff
@@ -1,16 +1,127 @@
...
+	devcontainer exec --workspace-folder . $(COMMAND)
+
+# Gems (and the rest of bin/setup) live inside the container, not in
+# anything DEVCONTAINER_STAMP can see, so this is tracked separately.
+# Depends on Gemfile/Gemfile.lock/bin/setup/config/database.yml.example -
+# editing any of them forces a recheck - and is cleared by dev-clobber's
+# `rm -rf .make` (a fresh bundler-cache volume wipes gems without touching
+# any of those files). If bin/setup fails - e.g. a freshly (re)created
+# volume defaults /bundle to root ownership - fix that and retry once
+# before giving up.
```
benrfairless, line 104: Stale: the bundler-cache volume was removed in `bea1fa1`, so there's no `/bundle` mount or cache volume left in `compose.yaml` (only `mysql-data`, `elasticsearch-data`, `mailpit-data`).

```suggestion
# `rm -rf .make` (a rebuilt container starts with no gems installed, without
# touching any of those files).
```

benrfairless (commented):
Reviewed at `8e12490` against the stack parent `feature/add-claude-md` (not `main`), so nothing from the parent PRs is credited or blamed here. CI is green at this exact SHA, so build/lint/test isn't an open question.

This is a big, careful piece of work and the docs carry it well - the memory table and the Bundler platform table both answer questions a new contributor would otherwise have to ask. `ENV.true?` is a genuine improvement over bare truthy checks, and the handling of where it *isn't* available (environments files falling back to `ActiveModel::Type::Boolean` with a comment saying why, plus the DECISIONS entry recording the exceptions) is the right level of documentation for a cross-cutting convention. `bin/setup-host`'s distro handling - the `UBUNTU_CODENAME` fallback, the `vm.max_map_count` check, the WSL2 note - will save people real time.

Inline comments cover the rest. Summary of what's not inline-able:

### One blocker: `Vagrantfile` is orphaned

Not in the diff, so I can't comment on the line. This PR deletes `local_dev/` entirely, but the tracked `Vagrantfile` still names `local_dev/ansible.cfg` and `local_dev/playbook.yml` as its `config_file` and `playbook` (lines ~107-108). `vagrant up` is now permanently broken with no way to recover from the repo.

The PR body says Vagrant was removed and the README's Vagrant section is gone, so deleting `Vagrantfile` looks like the intended fix and is a one-liner. Extra reason to delete rather than leave it: the file carries hardcoded dev passwords and `secret_key_base` values (pre-existing and intentional per its own inline warning), so removing it drops that from the tree too.

### PR description is stale on memory

The Resource Usage section still says the stack "fits within a ~6GB budget". After the RubyMine bump in `8e12490`, `rails-app` alone is `mem_limit: 8g`, total ceilings ~12.6GB. The README's memory table is already correct and explains ceilings-not-reservations well - this is a body edit only, no code change needed.

### Two questions

- **Merge order:** the body lists #1678 and #1675 as sitting before this one. Can you confirm the intended stack sequence?
- **CLA AI disclosure:** CLAUDE.md requires a note in the PR description for material AI involvement, distinct from commit trailers. Flagging as a check before this leaves draft - I can't tell from here whether it applies.

The MySQL and lockfile comments below need your answer rather than a patch; everything else has a suggested change attached.

---
*This review was drafted with AI assistance (Claude Code, claude-opus-5).*

README.md
```diff
@@ -1,160 +1,241 @@
...
 
-The original PHP app is also available at http://localhost:8080 but only if
-you're running an older branch (out of scope for this guide).
+If you want to run rails locally, but use the mysql, elasticsearch and mailpit services, then run:
 
-[1]: http://www.vagrantup.com/
-[2]: https://www.virtualbox.org/
-[3]: http://mailcatcher.me/
+```bash
+make dev-up
```
benrfairless, line 190: `make dev-up` starts the full stack including `rails-app`, which is the opposite of "just the other services". `make dev-hybrid` exists for exactly this case and its Makefile comment points straight back at this section.

```suggestion
make dev-hybrid
```

README.md
```diff
@@ -1,160 +1,241 @@
...
+design rationale):
+
+| System RAM | Expected experience                                                                                                                        |
+|------------|--------------------------------------------------------------------------------------------------------------------------------------------|
+| 8 GB       | Fine for a lightweight editor (e.g. vim), not a full IDE[^ide] like VS code and RubyMine                                                   |
+| 16 GB      | Comfortable for one project at a time alongside a full IDE[^ide], including via the full devcontainer. This is the baseline we design for. |
+| 32 GB      | Comfortable multitasking between two projects, or running the container stack alongside heavier tooling without adjustment.                |
+
+[^ide] full IDE: tested with VS code and RubyMine - which run a client in the devcontainer as well as on the host. The
+`ruby-app` container has a 8GB memory limit at RubyMine's recommendation. VS code needs less. IDEs use less memory in
```
benrfairless, line 106: Two small things in one line range:

1. A GitHub footnote definition needs the colon directly after the bracket (`[^ide]:`). As written this renders as literal text and the two `[^ide]` refs in the table above become broken links.
2. The service is `rails-app`, not `ruby-app` (matches `compose.yaml` and the rest of the README).

```suggestion
[^ide]: Full IDE - tested with VS code and RubyMine - which run a client in the devcontainer as well as on the host. The
`rails-app` container has a 8GB memory limit at RubyMine's recommendation. VS code needs less. IDEs use less memory in
```

docs/DECISIONS.md
```diff
@@ -5,6 +5,21 @@ wouldn't surface them. A decision local to one file/method belongs as a comment
 
 Append new entries at the top. Don't edit past entries except to mark them superseded (and say by what).
 
+## 2026-08-09 - Boolean ENV var flags use ENV.true?, not a bare truthy check
+
+Ruby only treats `nil`/`false` as falsy, so a bare `if ENV["FOO"]` treats any set value, including `"0"` and
+`"false"`, as true; `export FOO=""` in a shell sets an empty string rather than unsetting the variable, so blank
+counts as "set" too. `ENV.true?(key)` (`config/initializers/00_env_boolean.rb`, backed by
+`ActiveModel::Type::Boolean`) treats nil, blank, and false-like strings ("0", "false", "off", "f", case-insensitive)
```
benrfairless, line 13: "case-insensitive" isn't accurate, and it matters most here since this entry is meant to be the authoritative reference. `ActiveModel::Type::Boolean::FALSE_VALUES` lists only `"0"`, `"f"`, `"F"`, `"false"`, `"FALSE"`, `"off"`, `"OFF"` - so `BACKTRACE=False` or `=Off` evaluates **true**, which is exactly the surprise this decision is trying to remove.

```suggestion
`ActiveModel::Type::Boolean`) treats nil, blank, and false-like strings (exactly "0", "f", "F", "false", "FALSE",
"off", "OFF" - note mixed case like "False" is *not* one of them)
```

config/initializers/00_env_boolean.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Least-surprise boolean parsing for ENV var flags: nil, blank, and false-like values ("0", "false",
+# "off", "f", case-insensitive) are false; anything else is true. A bare `if ENV["FOO"]` treats any set
```
benrfairless, line 4: Same "case-insensitive" inaccuracy as the DECISIONS.md entry - `FALSE_VALUES` matches exact strings only, so `FOO=False` is true.

```suggestion
# Least-surprise boolean parsing for ENV var flags: nil, blank, and ActiveModel's exact false-like strings
# ("0", "f", "F", "false", "FALSE", "off", "OFF" - note "False" is not one of them) are false; anything else
# is true. A bare `if ENV["FOO"]` treats any set
```

bin/setup-host
```diff
@@ -0,0 +1,206 @@
...
+}
+
+install_docker_macos() {
+  if ! command -v brew >/dev/null 2>&1; then
+    echo "ERROR: Homebrew not found. Install it first: https://brew.sh" >&2
+    exit 1
+  fi
+
+  echo "Installing Docker Desktop via Homebrew..."
+  brew install --cask docker
```
benrfairless, line 126: Homebrew renamed this cask to `docker-desktop`. The old token still resolves (I confirmed `brew info --cask docker` redirects), so this works today - but pinning to the current name removes the dependency on that alias surviving, and `docker` is now also a formula (CLI only), which makes the old token ambiguous to read.

```suggestion
  brew install --cask docker-desktop
```

bin/setup-host
```diff
@@ -0,0 +1,206 @@
...
+install_devcontainer_cli() {
+  if ! command -v npm >/dev/null 2>&1; then
+    case "$(uname -s)" in
+      Linux)  install_node_linux ;;
+      Darwin) install_node_macos ;;
+    esac
+  fi
+
+  echo "Installing @devcontainers/cli..."
+  sudo npm install -g @devcontainers/cli
```
benrfairless, line 172: `sudo npm install -g` on macOS writes root-owned files into the Homebrew prefix, which then breaks later non-sudo `npm` operations for that user. Homebrew's node prefix is already user-writable so sudo isn't needed there. Worth splitting given macOS is the untested path on this PR.

```suggestion
  # Homebrew's node prefix is user-writable; sudo there leaves root-owned files
  # behind that break later non-sudo npm operations.
  if [ "$(uname -s)" = "Darwin" ]; then
    npm install -g @devcontainers/cli
  else
    sudo npm install -g @devcontainers/cli
  fi
```

ianheggie-oaf:
@benrfairless - I will flip the order back and put this first in a new stack as the previous PR was developed in conjunction with this one.

Issue #1270, opened 2021-10-26
[They Vote For You/production] ActionView::Template::Error: undefined method `possible_turnout' for nil:NilClass


mlandauer:
## Backtrace

    line 11 of [PROJECT_ROOT]/app/views/api/v1/divisions/_division.json.jbuilder: _app_views_api_v__divisions__division_json_jbuilder__3535966370156791926_70135259987100
    line 3 of [PROJECT_ROOT]/app/views/api/v1/divisions/index.json.jbuilder: _app_views_api_v__divisions_index_json_jbuilder___3089789612961555855_35066540

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/37542/faults/82232549)

Issue #1269, opened 2021-10-26
Delay between loading new divisions and division_info being populated


mlandauer:
## Backtrace

    line 14 of [PROJECT_ROOT]/app/models/division.rb: rescue in aye_majority
    line 14 of [PROJECT_ROOT]/app/models/division.rb: aye_majority
    line 74 of [PROJECT_ROOT]/app/models/division.rb: tied?

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/37542/faults/82182761)

mlandauer:
Since this causes the site to error for a couple of minutes each day when new divisions are loaded we should get this fixed soon. I guess it's not going to be super crucial to address for the upcoming election because once the election starts and people might refer to TVFY for reference parliament will not actually be sitting and so this problem shouldn't occur.

Taz17:
I thought about this, I came up with two ideas.

- Update `members cache` right after `application:load:members`, similarly update `divisions cache` right after `application:load:divisions` is executed. That reduces the time between `load` and `cache`.
This will require us update all the caches, excluding members and divisions since they have already been updated

- Update the cron job to run at an earlier time where the site receives less traffic.

Issue #1204, opened 2019-07-23
[They Vote For You/production] ActionView::Template::Error: Division#aye_majority delegated to division_info.aye_majority, but division_info is nil: #<Division id: 5762, valid: true, date: "2019-07-22", number: 7, house: "senate", name: "Business &#8212; Rearrangement", source_url: "http://parlinfo.aph.gov.au/parlInfo/search/display...", debate_url: "http://parlinfo.aph.gov.au/parlInfo/search/display...", motion: "<p class=\"speaker\">Penny Wong</p>\n\n<p>I understand...", clock_time: "021:02:00", source...


henare:
## Backtrace

    line 11 of [PROJECT_ROOT]/app/models/division.rb: rescue in aye_majority
    line 11 of [PROJECT_ROOT]/app/models/division.rb: aye_majority
    line 71 of [PROJECT_ROOT]/app/models/division.rb: tied?

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/37542/faults/33534926)

jamezpolley:
https://theyvoteforyou.org.au/divisions/senate/2019-07-22/7

This seems to be a one-off; I can't reproduce it on either https://theyvoteforyou.org.au/people/senate/nsw/concetta_fierravanti-wells or https://theyvoteforyou.org.au/divisions/senate/2019-07-22/7

The timing seems to be around the time we'd be doing the daily update, so it may have just been a page view that happened just as this division was being updated.

PR #1697, opened 2026-08-18
Build division caches as divisions are loaded

### LINKED ISSUE #1641: [They Vote For You/production] NameError: undefined local variable or method 'aye_majority' for nil
**Describe the bug**
Whenever we run `RAILS_ENV=production bundle exec rake "application:load:divisions[2026-05-01,2026-06-17]"` it breaks the divisions on the site.

**To Reproduce**
1. Run the above command
2. Attempt to browse to `https://theyvoteforyou.org.au/divisions/all`

**Expected behavior**
- A list of divisions should appear, and no errors should be generated in Honeybadger.

**Screenshots/Output**
Backtrace
    line 18 of [PROJECT_ROOT]/app/models/division.rb: Division#aye_majority
    line 86 of [PROJECT_ROOT]/app/models/division.rb: Division#tied?
    line 81 of [PROJECT_ROOT]/app/models/division.rb: Division#passed?

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/37542/faults/123698180)

**Additional context**
The `application:load:divisions` task only writes `Division` and `Vote` rows — it never builds the cache tables the division pages render from. The pages read aggregate values (`turnout`, `aye_majority`, `majority`, `rebellions`, etc.) via methods that `Division` delegates to its `division_info` association (`app/models/division.rb:18`), with no `allow_nil`. Freshly loaded divisions have no `division_info` row, so the delegation raises `NameError`/`Module::DelegationError` and every division page 500s. The whip summary partials similarly depend on the `Whip` cache.

Render path that triggers the crash:
- `app/views/divisions/_summary.html.haml:6` → `division_outcome_with_majority_strength`
- → `app/helpers/divisions_helper.rb:97` `division_outcome` → `division.passed?`
- → `division.rb:81` `passed?` → `tied?` → `aye_majority` (delegated to a nil `division_info`)

The `DivisionInfo` rows are populated by a separate step, `DivisionInfo.update_all!`, run by the `application:cache:division` task (`lib/tasks/application.rake:38`); whips are built by `Whip.update_all!` in `application:cache:whip`. The load task runs neither. This is why the `daily` task (`lib/tasks/application.rake:103-110`) always follows the load with `application:cache:all` — running the bare load command skips that cache rebuild.

**Workaround:** run a cache rebuild after loading, e.g. `RAILS_ENV=production bundle exec rake application:cache:all` (or `application:cache:all_except_people_distances` to skip the slow people-distances step; `application:cache:division` is the minimum needed to clear the 500s).

**Possible fixes:** have the load task rebuild the relevant caches itself, and/or add `allow_nil: true` to the `division_info` delegation so a missing cache degrades gracefully instead of raising.

benrfairless:
## Description

Divisions were loaded by `application:load:divisions`, but their `division_info` and `whip` cache rows weren't built until the separate `application:cache:all` step. `Division` delegates `turnout`, `aye_majority`, `rebellions`, `majority` and `majority_fraction` to `division_info` with no `allow_nil`, so in between those two steps every page that renders a division returned a 500.

The loader wraps **each division** in its own transaction, so a division became visible to web traffic as soon as that transaction committed. The window therefore opened mid-load, and the daily 9:15am cron hit it just as a manual load did.

Two layers, matching the two fixes suggested on #1641:

**Close the window.** Both caches are now built inside the loader's per-division transaction, so a division and its caches become visible together. Whips are built first, because `DivisionInfo`'s rebellion counts derive from them (`Vote.rebellious` inner joins `whips`, which is why the rake file already declares `task division: :whip`). `DivisionInfo.update_divisions!` and `Whip.update_divisions!` take the ids to rebuild, or `nil` for all of them, which is what `update_all!` now passes. `application:load:daily` and `application:cache:all` are unchanged.

**Degrade instead of crashing.** If a cache row is missing for any other reason, the delegation now allows nil - matching the existing `member_info` delegation on `Member` - and the outcome renders as "Unknown" rather than asserting a result we haven't calculated. Showing "Not passed" for a division that may well have passed would misreport a parliamentary vote, so the helpers check `outcome_known?` before displaying an outcome. `allow_nil` alone wouldn't have been enough for the API: `api/v1/divisions/_division.json.jbuilder` reads `division_info` directly rather than through the delegation, so it gets its own guard and emits `null`.

Also switches `DivisionInfo`'s save to `update!` - a silently failed save was one of the ways a division could end up with no cache row.

Not addressed here: `MemberInfo` has the identical load/cache gap, but `Member`'s delegation is already guarded so it degrades quietly rather than 500ing. Worth a separate issue.

## Motivation and Context

Resolves #1641.

Sentry recorded a burst of this on 2026-08-18, 23:16:34-23:22:53 UTC - about 90 seconds after 09:15 AEST, matching the `15 9 * * 1-5` cron. ~36 events across five issues:

| Sentry | Controller | Events |
| --- | --- | --- |
| THEY-VOTE-FOR-YOU-A | `Api::V1::DivisionsController#index` | 23 |
| THEY-VOTE-FOR-YOU-E | `MembersController#show` | 7 |
| THEY-VOTE-FOR-YOU-C | `DivisionsController#index` | 4 |
| THEY-VOTE-FOR-YOU-D | `HomeController#search` | 1 |
| THEY-VOTE-FOR-YOU-B | `PeopleDistancesController#show` | 1 |

This is long-standing: #1269 described the same daily-cron window back in 2021 and was closed as stale; #1204 and #1270 are earlier duplicates.

## How Has This Been Tested?

- [x] Ran automated tests on my own system
- [x] Confirmed it passed the GitHub actions tests

448 examples, 0 failures. `rubocop --parallel` clean.

Every layer is covered by a test that was confirmed to **fail without its fix**:

- `spec/lib/data_loader/debates_spec.rb` (new) - drives the real loader against the existing `2009-11-25.xml` fixture with the HTTP fetch stubbed, and asserts each loaded division comes out with its caches built. 5 of its 7 examples fail without the loader change.
- `spec/requests/divisions_controller_uncached_spec.rb` (new) - renders `/divisions/all` with an uncached division, asserts 200 and that the page says "Unknown" rather than claiming an outcome. Fails with `NoMethodError` without `allow_nil`.
- `spec/requests/api/v1/divisions_controller_spec.rb` (new) - the regression test for THEY-VOTE-FOR-YOU-A. The API namespace had no coverage at all before this.
- `spec/models/division_spec.rb` - `tied?` and `unanimous?` had no direct coverage previously.
- `spec/models/division_info_spec.rb`, `spec/models/whip_spec.rb` - the scoped rebuilds, including that each produces the same result as a full rebuild.

Two things I checked specifically:

1. **No fixture churn.** `spec/html_compare_helper.rb` hardcodes `overwrite = true`, so a rendering change silently rewrites the static-page fixtures. `spec/fixtures/` is untouched after a full run, as expected - the fixture divisions all have a `division_info`, so the Unknown branch never fires.
2. **No performance regression on the nightly rebuild.** Having `update_all!` pass every division id would have produced an `IN (...)` list with tens of thousands of entries. It passes `nil` instead, so the full rebuild emits exactly the same unscoped aggregates as before - verified by capturing the SQL. Divisions with no votes at all still get their row of zeroes.

## Types of Changes

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

Assisted-by: Claude Code:claude-opus-5

### DIFF
--- a/app/controllers/divisions_controller.rb
+++ b/app/controllers/divisions_controller.rb
@@ -76,7 +76,7 @@ class DivisionsController < ApplicationController
     if @division.nil?
       render "home/error404", status: :not_found
     else
-      @rebellions = @division.votes.rebellious.order("members.last_name", "members.first_name") if @division.rebellions.positive?
+      @rebellions = @division.votes.rebellious.order("members.last_name", "members.first_name") if @division.rebellions.to_i.positive?
       @whips = @division.whips.order(:party)
       @votes = @division.votes.joins(:member).includes(:member).order("members.party", "vote", "members.last_name", "members.first_name")
 
--- a/app/helpers/divisions_helper.rb
+++ b/app/helpers/divisions_helper.rb
@@ -64,6 +64,8 @@ module DivisionsHelper
   end
 
   def majority_strength_in_words(division)
+    return "" unless division.outcome_known?
+
     if division.unanimous?
       "unanimously"
     elsif division.tied?
@@ -94,10 +96,14 @@ module DivisionsHelper
 
   # TODO: We should be taking into account the strange rules about tied votes in the Senate
   def division_outcome(division)
+    return "Unknown" unless division.outcome_known?
+
     division.passed? ? "Passed" : "Not passed"
   end
 
   def division_outcome_class(division)
+    return "division-outcome-unknown" unless division.outcome_known?
+
     division.passed? ? "division-outcome-passed" : "division-outcome-not-passed"
   end
 
--- a/app/lib/data_loader/debates.rb
+++ b/app/lib/data_loader/debates.rb
@@ -53,6 +53,12 @@ module DataLoader
 
                 Vote.create!(division: division, member: member, vote: vote[0], teller: vote[1])
               end
+
+              # Build the caches the division pages read from before we commit, so
+              # the division is never visible without them. Whips first - the
+              # rebellion counts in DivisionInfo are derived from them.
+              Whip.update_divisions!(division.id)
+              DivisionInfo.update_divisions!(division.id)
             end
           end
         end
--- a/app/models/division.rb
+++ b/app/models/division.rb
@@ -15,7 +15,7 @@ class Division < ApplicationRecord
   has_many :wiki_motions, -> { order(created_at: :desc) }, inverse_of: :division, dependent: :destroy
   has_and_belongs_to_many :bills
 
-  delegate :turnout, :aye_majority, :rebellions, :majority, :majority_fraction, to: :division_info
+  delegate :turnout, :aye_majority, :rebellions, :majority, :majority_fraction, to: :division_info, allow_nil: true
 
   scope :in_date_range, ->(date_start, date_end) { where(date: date_start...date_end) }
   scope :in_house, ->(house) { where(house: house) }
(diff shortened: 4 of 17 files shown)


ianheggie-oaf (approved):
Approved- lgtm - reasoning is sound, changes protected by tests, reminded myself that nil.to_I is 0 so .positive? Test works.

PR #1683, opened 2026-08-11
Add Sentry error monitoring alongside Honeybadger

benrfairless:
## Description

Adds Sentry error monitoring, running in parallel with Honeybadger for a verification period before Honeybadger is removed in a follow-up:

- **Backend errors + tracing**: `sentry-ruby`/`sentry-rails` with a 10% traces sample rate, breadcrumbs from logs/HTTP, `send_default_pii` off. DSN comes from encrypted credentials (separate client keys for production and staging, so each can be rate-limited or deactivated independently), so the SDK is dormant in development and test.
- **Browser JavaScript**: Sentry Loader Script partial rendered alongside the existing Honeybadger snippet, tagged with the same environment and release as the backend.
- **Cron monitoring**: `application:load:daily` and `application:cron:nightly` send check-ins; monitors are auto-created in Sentry on first production run with the schedules from the infrastructure repo (Australia/Sydney).
- **Release/deploy tracking**: a Capistrano hook runs `sentry-cli` on the deployer's machine after publishing — releases are named by the full git SHA, matching the SDK's auto-detected release from Capistrano's `REVISION` file. Deploys without sentry-cli set up still work; the step is skipped with a warning. Setup instructions are in the README.

Out of scope for now: JS source maps (Sprockets/terser produces none), session replay, logging, profiling.

## Motivation and Context

We constantly hit monthly limits with Honeybadger, and they no longer offer an OSS plan. Sentry has given us a sponsored Business plan. Part of #1607 (kept open — per the issue, both services run in parallel until we've verified Sentry, then Honeybadger removal happens in a follow-up PR).

After merge:

1. Deploy to staging, trigger an error, and confirm it appears in **both** Honeybadger and Sentry with matching release/deploy records
2. Confirm the first production cron runs show green check-ins in [Crons](https://oaf-org-au.sentry.io/insights/crons/)
3. Link the GitHub integration to this repo in Sentry (Settings → Integrations → GitHub) to get suspect-commit PR links; commit association already works via `--local`

## How Has This Been Tested?

- [x] Checked affected area manually on my own / staging system
  - A real exception was sent through the SDK and confirmed in Sentry with a readable stack trace and correct environment/release tags ([THEY-VOTE-FOR-YOU-1](https://oaf-org-au.sentry.io/issues/THEY-VOTE-FOR-YOU-1), resolved as a deliberate test)
  - `cap -T` lists `sentry:notify_deploy`; the guard path warns and continues (exit 0) when sentry-cli can't run
[...]

PR #1701, opened 2026-08-18
Extend Sentry: CLI v4 support, PII, trace stitching, cron instrumentation

benrfairless:
## Description

Three related extensions to our Sentry setup:

1. **Sentry CLI v4 support in deploy release tracking** (`lib/capistrano/tasks/sentry.rake`). CLI v4 renamed the binary (`sentry-cli` → `sentry`), the `releases` command group (→ `release`), and replaced `deploys new` with `release deploy` ([migration guide](https://cli.sentry.dev/migrating-from-v3/)). The Capistrano task now detects whichever binary is installed, preferring v4, so deployers can migrate at their own pace. It also fixes the silent failure that left release `e4db521` orphaned in Sentry with no commits or deploy: the four CLI steps shared one `rescue`, so a `set-commits` failure skipped `finalize` and the deploy record. Each step now warns and continues independently, commits are associated server-side via the GitHub integration (`--commit repo@sha`) with `--local` as fallback, and org/project are passed explicitly (v4 doesn't read the project from the repo `.sentryclirc` on `release new`). README setup instructions updated for both versions.

2. **PII + user context + trace stitching**. `send_default_pii` enabled on both backend and browser so events carry IPs and request data; signed-in users attached to events (id/email/name) on both sides. `filter_parameters` extended to scrub the API `?key=` query parameter and Devise reset/confirmation tokens from captured request data (and from Rails logs, where they were previously exposed). Sentry trace-propagation meta tags rendered in the layout so browser pageload transactions continue the Rails trace.

3. **Spans + metrics on the cron pipelines**. `with_sentry_cron_monitoring` now starts a transaction per cron run (`sampled: true` — one run a day would mostly be dropped at our 0.1 sample rate), and a `sentry_stage` helper wraps each stage of the daily load and nightly cards runs, so traces show where the hours go. `DataLoader::Debates` gets per-fetch and per-division spans plus counters (`data_load.divisions.loaded`/`.mismatch`/`.xml_missing` by house — the reload mismatch was previously only a log warning), a `data_load.members.total` gauge, and a `cards.screenshots.generated` counter.

## Motivation and Context

Follows on from #1683/#1689 (Sentry adoption, Honeybadger removal) and #1700. The release-tracking half was diagnosed live: the current production release `e4db521` sat in Sentry unfinalized with zero commits and no deploy record because of the swallowed `set-commits` failure. That release has been backfilled by hand (commits, finalize, production deploy with real timestamps); this PR stops it happening again and makes the tracking work with the new CLI.

## How Has This Been Tested?

  - The Capistrano task's exact v4 command sequence was run end to end against the real Sentry org using a probe release (create → set-commits via GitHub integration → finalize → deploy) and then deleted. Both CLI detection branches exercised.
  - Span/metric behaviour smoke-tested in-process against the bundled sentry-ruby 6.7.0: uninitialized → safe no-op (dev has no DSN); initialized → cron transaction → nested stage/child spans + trace-linked metric all created.
  - RuboCop clean on all touched files; Haml partial compiles; full Rakefile loads (`rake -T` resolves the tasks).

- [ ] Ran automated tests on my own system

  Local checkout has no `config/database.yml`, so the suite runs in CI only.

- [ ] Confirmed it passed the GitHub actions tests

**Still needs confirming after deploy:** next `application:load:daily` run should produce a trace named `application-load-daily` with the stage breakdown and `data_load.*` metrics; a signed-in error event should carry user context; page source should show the `sentry-trace`/`baggage` meta tags; the next deploy should attach commits + deploy record to its release.

## Types of Changes

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation

## Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,13 +179,20 @@ Each deploy notifies [Sentry](https://oaf-org-au.sentry.io/) of the new
 release so that errors can be linked back to the commits that introduced them.
 This runs on your machine as part of the deploy and needs a one-off setup:
 
-1. Install [sentry-cli](https://docs.sentry.io/product/cli/):
-   `brew install getsentry/tools/sentry-cli` on a Mac, or on Linux
-   `curl -sL https://sentry.io/get-cli/ | sh`.
-2. Create a personal auth token at https://oaf-org-au.sentry.io under
-   Settings → Auth Tokens → Create New Token, with only the
-   `project:releases` scope.
-3. Put the token in `~/.sentryclirc`:
+1. Install the [Sentry CLI](https://cli.sentry.dev/getting-started/) (v4,
+   the `sentry` binary): `brew install getsentry/tools/sentry` on a Mac, or
+   on Linux `curl https://cli.sentry.dev/install -fsS | bash`. The older v3
+   `sentry-cli` binary is also still supported if you have it installed.
+2. Authenticate. The easiest way is the browser-based login (v4 only):
+
+   ```
+   sentry auth login
+   ```
+
+   Alternatively (and for v3), create a personal auth token at
+   https://oaf-org-au.sentry.io under Settings → Auth Tokens → Create New
+   Token, with only the `project:releases` scope, and put it in
+   `~/.sentryclirc`:
 
    ```
    [auth]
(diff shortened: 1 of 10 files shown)


ianheggie-oaf (approved):
Approved - Given this has come out of production experience and that it will be checked after merge

PR #1651, opened 2026-07-06
Upgrade Ruby to 3.4.10

ianheggie-oaf:
## What and why

Closes #1651's own goal: latest 3.4.x for security updates. This branch had gone stale (based on an old `main` - merging it as-is would have reverted the recent MySQL 8.4 CI fix, deleted `.github/copilot-instructions.md`, and downgraded a couple of GitHub Action versions), so I reset it cleanly onto current `main` and reapplied just the actual upgrade.

Went with **3.4.10** rather than the 3.4.9 this issue originally asked for - the latest available patch has moved on since it was filed, and 3.4.10 (released 30 June 2026) bundles a net-imap security fix, which better matches the issue's own stated goal of "brings in latest security updates." Also bumped `mise.toml`'s pin to match, since it's meant to stay in sync with `.ruby-version`.

**A real prerequisite found along the way:** bootsnap 1.18.6 fails under Ruby 3.4.10 with `RuntimeError: should not compile with coverage` as soon as SimpleCov is active - every single spec file errors before it can load, 0 examples run. Bumped it to 1.26.0, which fixes it.

## Type of change

- [x] Bug fix

## How this was checked

- [x] Ran the full test suite: 475 examples, 1 failure - a pre-existing, unrelated Selenium/Chrome sandbox issue (`spec/features/edit_division_spec.rb:13`), not a regression from this change
- [x] Ran `bundle exec rubocop --parallel` across the whole app - clean
- [ ] Checked the affected area on my own or a staging system
- [ ] Ran `/code-review` (or equivalent): fixed any concerns (and rechecked), linked follow-up issues below, or noted why they are not important below
- [ ] GitHub Actions tests: likewise fixed so they pass or linked a follow-up issue / replied why not important directly on each comment (reviewer will mark resolved as part of review)
- [ ] Documentation updated, or not needed

Assisted-by: Claude Code:claude-sonnet-5

PR #1756, opened 2026-09-14
Upgrade Ruby version to 3.4.10

Br3nda:
replaces:
 - #1651 

## What and why
Upgrades Ruby to current stable 3.x series

## Type of change

- [ ] Chore

## How this was checked

tested on dev. 

I used the freerange model in my brain to make this.


benrfairless (commented):
From AI review:

>Production deploy will fail unless RVM on the server already has 3.4.10. config/deploy.rb:6 now derives rvm_ruby_version from .ruby-version, so the first cap production deploy after merge will rvm use 3.4.10 on the box. Nothing in this repo installs it. Worth confirming the server side (or the openaustralia/infrastructure Ansible run) before merging, and saying so in the PR description. Same applies to bin/passenger-ruby:3, which also rvm uses from .ruby-version.

You will need to raise a PR in infrastructure to update the ruby version before this lands.

benrfairless (approved):

